PR #8, opened 2026-04-01
:recycle: Utilize new QDMI functionalities available in the `1.2.x` release

marcelwa:
This PR bumps QDMI to the latest `1.2.x` commit and utilizes its new functionality as specified in its [upgrading guide](https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/c5f71c63d5bf3b97c26c60efd367a4be3670d3a0/UPGRADING.md).


docs/dependencies.md
burgholzer, line 1: The QDMI version needs a bump here as well

src/CMakeLists.txt
```diff
@@ -30,8 +30,13 @@ target_compile_features(${QDMI_TARGET_NAME} PRIVATE cxx_std_20)
 target_compile_definitions(${QDMI_TARGET_NAME}
                            PRIVATE QDMI_VERSION="${QDMI_VERSION}")
 target_compile_definitions(
```
burgholzer, line 30: This can be combined with the line right above to set all compile definitions in one command.

src/CMakeLists.txt
```diff
@@ -53,23 +58,28 @@ target_sources(
          FILES
          ${PUBLIC_HEADER_FILES})
 
+if(TARGET qdmi::qdmi)
+  get_target_property(QDMI_INCLUDE_DIRS qdmi::qdmi
+                      INTERFACE_INCLUDE_DIRECTORIES)
+  if(QDMI_INCLUDE_DIRS)
+    target_include_directories(${QDMI_TARGET_NAME} PUBLIC ${QDMI_INCLUDE_DIRS})
+  endif()
+endif()
```
burgholzer: This can't be right. The target should not depend on the QDMI headers as all headers are vendored as part of the device setup.
This code should be removed.

test/integration/CMakeLists.txt
```diff
@@ -15,15 +15,15 @@
 #
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-add_executable(integration_tests test_iqm_device_integration.cpp)
+add_executable(integration_tests ${CMAKE_SOURCE_DIR}/src/logging.cpp
```
burgholzer: Why wasn't this necessary before?

test/unit/CMakeLists.txt
```diff
@@ -15,18 +15,33 @@
...
 
-add_executable(unit_tests test_iqm_api_config.cpp test_iqm_auth.cpp
-                          test_iqm_device.cpp test_logging.cpp)
+add_executable(
+  unit_tests
+  ${CMAKE_SOURCE_DIR}/src/curl_http_client.cpp
+  ${CMAKE_SOURCE_DIR}/src/iqm_api_config.cpp
+  ${CMAKE_SOURCE_DIR}/src/iqm_auth.cpp
+  ${CMAKE_SOURCE_DIR}/src/iqm_device.cpp
+  ${CMAKE_SOURCE_DIR}/src/logging.cpp
```
burgholzer: This has not been necessary before. Why is it now?

test/unit/CMakeLists.txt
```diff
@@ -15,18 +15,33 @@
...
+  test_iqm_auth.cpp
+  test_iqm_device.cpp
+  test_logging.cpp)
 target_compile_features(unit_tests PRIVATE cxx_std_20)
-target_include_directories(unit_tests PRIVATE ${CMAKE_SOURCE_DIR}/include)
+target_compile_definitions(
+  unit_tests
+  PRIVATE QDMI_VERSION="${QDMI_VERSION}"
+          IQM_QDMI_DEVICE_VERSION="${CMAKE_PROJECT_VERSION}"
+          IQM_QDMI_device_EXPORTS)
```
burgholzer: This is wrong and must only be set on the library itself.

test/unit/CMakeLists.txt
```diff
@@ -15,18 +15,33 @@
...
+                                              ${CMAKE_BINARY_DIR}/src/include)
 target_link_libraries(
   unit_tests
   PRIVATE gtest_main
           gmock_main
           gtest
           gmock
-          qdmi::qdmi
           qdmi::qdmi_project_warnings
-          ${QDMI_TARGET_NAME}
```
burgholzer, line 30: Presumably, this is the cause of the other changes. I don't quite understand this change. Could you elaborate?

burgholzer (changes requested):
Thanks @marcelwa 🙏
Please find some feedback attached. Hopefully that makes sense.

src/CMakeLists.txt
marcelwa: Tbh, I got really confused by the [QDMI Upgrading Guide](https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/v1.2.x/UPGRADING.md#new-header-for-managing-exported-symbols-of-device-implementations). This explanation also applies to the other changes that you flagged. I misunderstood the whole signal visibility and macro definition. I think I got it now.

include/curl_http_client.hpp
burgholzer, line 1: Why do these classes (including many in the other files) need to be exported?
They are not part of the public API of the device and, therefore, should not be exposed in general.
My best guess is that this is due to testing, but even then, this should not be necessary.
Potentially we want to split the curl/auth/logging code into separate targets that can be linked to; and that are built as static libraries. Then these could be used in the tests.
But this is mostly guesswork.

The QDMI functions that are automatically exposed through the prefixed header should be the only exported symbols of the shared library.

src/iqm_device.cpp
```diff
@@ -582,7 +582,7 @@ int IQM_QDMI_device_update_dynamic_quantum_architecture(
 } // namespace
 
 // NOLINTNEXTLINE(misc-use-internal-linkage)
-int IQM_QDMI_device_session_init_with_http_client(
+IQM_QDMI_EXPORT int IQM_QDMI_device_session_init_with_http_client(
```
burgholzer, line 585:
```suggestion
int IQM_QDMI_device_session_init_with_http_client(
```
This is a function in a source file. I fail to see how this would need an export.

test/unit/CMakeLists.txt
```diff
@@ -18,14 +18,14 @@
 add_executable(unit_tests test_iqm_api_config.cpp test_iqm_auth.cpp
                           test_iqm_device.cpp test_logging.cpp)
 target_compile_features(unit_tests PRIVATE cxx_std_20)
-target_include_directories(unit_tests PRIVATE ${CMAKE_SOURCE_DIR}/include)
+target_include_directories(unit_tests PRIVATE ${CMAKE_SOURCE_DIR}/include
+                                              ${CMAKE_BINARY_DIR}/src/include)
```
burgholzer:
```suggestion
target_include_directories(unit_tests PRIVATE ${CMAKE_SOURCE_DIR}/include)
```
This change doesn't quite feel right. We are still linking against the device here, which should pull the include directory from the build tree in.
At the same time, this probably will need some adjustments based on the other bigger comment.

burgholzer (changes requested):
Thanks @marcelwa for another round of iteration!
I think we are getting there. However, the symbol visibility still needs a bit of work.

src/iqm_device.cpp
marcelwa: Because we use it in the unit tests here: https://github.com/iqm-finland/QDMI-on-IQM/blob/main/test/unit/test_iqm_device.cpp#L42-L44 Without the export, the symbol cannot get resolved.

include/curl_http_client.hpp
marcelwa: Yes, it is indeed due to testing. Without the exposure, the tests cannot compile. The alternative would indeed be to refactor the utility code into a separate target. I can make a proposal for that.

burgholzer: Yeah, that would be good. We shouldn't let testing dictate the cleanliness of the shared library we produce.
Compiling the utility code into a static library that can also be linked in the tests seems reasonable.

src/CMakeLists.txt
burgholzer: If you do not mind, would you be open to create a PR in the QDMI repository that updates the respective section in the upgrade guide to improve its clarity?

docs/dependencies.md
burgholzer: Done in https://github.com/iqm-finland/QDMI-on-IQM/pull/8/commits/9363bacdd42ddf27c74d6fe2fb2e528c172a3ac7

burgholzer (approved):
This is ready to merge now.
I cleaned up a lot of the CMake configuration from the PR and enabled all but the mock tests for the installed mode.

PR #11, opened 2026-04-13
♻️ Return device alias as `QDMI_DEVICE_PROPERTY_NAME` instead of static `"IQM QDMI Device"`

marcelwa:
This PR refactors the behavior of `QDMI_DEVICE_PROPERTY_NAME`. It now returns the quantum computer alias of the current session instead of the static `"IQM QDMI Device"` string.


burgholzer (approved):
Nice! LGTM 🚀

PR #37, opened 2026-04-24
✨ Add Qiskit-compatible `IQMBackend` and Sampler and Estimator primitives using MQT Core

marcelwa:
This PR adds MQT Core as a dependency to enable a Qiskit-compatible workflow. It adds an `IQMBackend` as well as Sampler and Estimator primitives that can be accessed via respective `.sampler()` and `.estimator()` member functions.


burgholzer (approved):
Thanks for your initial work on this @marcelwa 🙏🏼 
I pushed some changes that address various aspects of the PR that weren't quite where we would want them to be. I hope the individual commits are clear enough.

For the most part I:
- turned this into an optional dependency because not everyone will want to interact with the QDMI device via Qiskit. Also added it as a dependency group because developers will always want to at least be able to run the respective tests. Also widened the range of supported versions and made the supported Qiskit versions more explicit.
- simplified/fixed configuration code for pytest and cibuildwheel as well as the test setup itself.
- actually enabled the tests to run in CI, which revealed a subtle issue with the Windows path logic. The respective fix needs to be backported to the QDMI template.
- turned the docs into interactive ones that actually execute something. Reworded parts of the documentation and made sure to properly cross-reference where applicable so the docs render nicely.

This should now be good to go; but please have another look.

There is one more thing I noticed, that I would want to get rid of rather quickly: Since all of the testing we do goes through Resonance, every single call that initiates a QDMI session will trigger an error warning in the logs from the failed attempt to access the calibration endpoint. This is annoying and confusing to users, as it does not represent an actual failure case. We should find a way to silence this particular error so that it is not constantly raised. That's for an immediate follow-up though.

marcelwa:
Many thanks for the changes and additions, @burgholzer! :pray: 

The QDMI device template and example device path resolution fix is in https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/409

The warning level demotion of missing endpoints in https://github.com/iqm-finland/QDMI-on-IQM/pull/51

burgholzer:
Nice. Many thanks 🙏

PR #51, opened 2026-04-28
🚸 Demote warnings from missing calibration enpoints to debug level

marcelwa:
Since the integration tests go through Resonance, every call that initiates a QDMI session logs a warning about the failed attempt to access the calibration endpoint. This was annoying and confusing for users, as it did not represent a real failure case. This PR demotes such warnings to debug level.


burgholzer (approved):
Nice. Glad to see this warning gone 🙌

Edit: @marcelwa just saw that this does not currently build. It looked good though in general. So feel free to merge once this builds.

PR #52, opened 2026-04-29
🚸 Add explicit retry logic to avoid hitting API rate limits

marcelwa:
This pull request introduces several improvements to the `CurlHttpClient` implementation, focusing on enhanced testability, stricter SSL verification, and improved handling of HTTP 429 rate limiting. The changes also refactor internal logic into a new internal header, update the test infrastructure, and improve the CMake test configuration.

**Core improvements and refactoring:**

* Extracted internal helpers, error handling policies, and retry logic from `curl_http_client.cpp` into a new internal header `curl_http_client_internal.hpp` for better encapsulation and testability. [[1]](diffhunk://#diff-94cd9936562fe8b9ee4acc4887ecb22f72a72fcec5c0f6e50b703d5f82c409c3R1-R171) [[2]](diffhunk://#diff-bff8c79a2b98b96ef00098fd7dcf5ac93cfdf7f99e32c28b957c765e91e978d2L38-L106)
* Added a new test support file `curl_http_client_test_support.cpp` with functions to exercise internal logic, such as response code handling and retry behavior, without requiring live network connections.

**Security and reliability enhancements:**

* Enabled strict SSL certificate and hostname verification by setting `CURLOPT_SSL_VERIFYPEER` to `1L` and `CURLOPT_SSL_VERIFYHOST` to `2L` in all HTTP requests. [[1]](diffhunk://#diff-cf982d07261e893424c6bb56fc541e0651efbd84c343f6a450a65ba7b62305a1L42-R43) [[2]](diffhunk://#diff-bff8c79a2b98b96ef00098fd7dcf5ac93cfdf7f99e32c28b957c765e91e978d2L277-R266) [[3]](diffhunk://#diff-bff8c79a2b98b96ef00098fd7dcf5ac93cfdf7f99e32c28b957c765e91e978d2L337-R299)
* Implemented automatic retry logic for HTTP 429 ("Too Many Requests") responses, with configurable retry count and delay. [[1]](diffhunk://#diff-cf982d07261e893424c6bb56fc541e0651efbd84c343f6a450a65ba7b62305a1L42-R43) [[2]](diffhunk://#diff-94cd9936562fe8b9ee4acc4887ecb22f72a72fcec5c0f6e50b703d5f82c409c3R1-R171)

**Build and test infrastructure updates:**

* Updated the unit test CMake configuration to include the new test support file and link against `CURL::libcurl`.
* Minor workflow changes: reordered arguments for `ctest` invocations in GitHub Actions workflows to avoid ambiguity, and clarified comments. [[1]](diffhunk://#diff-eaa888fc3a429f7ca944f775f639188c93e29cd8e33e22b145d2071fa6a6604cL79-R79) [[2]](diffhunk://#diff-f2cdba000bb15b1e3a9d6509cf4020074c43273f77aaf415da609f5d2aedbfdaL94-R94) [[3]](diffhunk://#diff-097de72b3ca6a8789796768d3d4328f7646773fd09799ced21b9a62092528adbL53-R53)

These changes collectively improve the maintainability, security, and test coverage of the HTTP client component.


burgholzer (changes requested):
I believe I would have went a different way with this.
We can already intercept the rate limit error code at the HTTP request level withing the QDMI device implementation itself and can simply re-issue the request after a certain delay.
Then the CI and the tooling around it can practically stay the same.

marcelwa:
@burgholzer your proposed way is now implemented. We can compare the two. I'm in favor of my initial attempt since it is much simpler. Curious to hear your thoughts, though :)

burgholzer:
> @burgholzer your proposed way is now implemented. We can compare the two. I'm in favor of my initial attempt since it is much simpler. Curious to hear your thoughts, though :)

In my opinion, the first solution in this PR maskes/hides a problem, while the second solution actually solves the underlying problem. Although I cannot help but feel that the current solution feels very bloated and quite a bit over the top for a simple retry mechanism. I do not have the capacity at the moment to trial this myself. However, I feel there is a solution to this that is much more compact and straightforward.

marcelwa:
@burgholzer I tried to simplify the mechanism. If Teemu cannot get us a token without the limit, we could come back to this PR.

src/internal/curl_http_client.cpp
```diff
@@ -24,16 +24,20 @@
 
 #include "curl_http_client.hpp"
 
+#include "../../test/unit/curl_http_client_test_support.hpp"
```
burgholzer: Includes like these are a no-go. Why would the _internal source code_ include something from the test tree?

src/internal/curl_http_client.cpp
```diff
@@ -52,6 +56,43 @@ enum class ERROR_LOG_POLICY : uint8_t {
...
+}
+
+/**
+ * @brief Sleep for the given number of seconds before retrying a request.
+ *
+ * @param seconds The retry delay in seconds.
+ */
+void Sleep_for_seconds(const uint8_t seconds) {
+  std::this_thread::sleep_for(std::chrono::seconds(seconds));
+}
```
burgholzer: This is 8 lines of code for a one-liner. Is this really necessary/helpful?

src/internal/curl_http_client.cpp
```diff
@@ -274,18 +345,21 @@ int Perform_get_request(const std::string &url, const std::string &bearer_token,
   curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, Curl_write_callback);
   curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1L);
   curl_easy_setopt(curl, CURLOPT_TIMEOUT, 3600L);
-  curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, 0L);
-  curl_easy_setopt(curl, CURLOPT_SSL_VERIFYHOST, 0L);
+  curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, 1L);
+  curl_easy_setopt(curl, CURLOPT_SSL_VERIFYHOST, 2L);
```
burgholzer, line 250: What are these changes for? Same question applies below.

src/internal/curl_http_client.cpp
```diff
@@ -274,18 +345,21 @@ int Perform_get_request(const std::string &url, const std::string &bearer_token,
...
-  const auto ret = Handle_response_code(curl, url, response, error_log_policy);
+  const auto ret = Perform_request_with_retries(
+      url, response, error_log_policy,
+      [&]() -> Request_attempt_result {
+        const auto res = curl_api_hooks.easy_perform(curl);
+        if (res != CURLE_OK) {
+          return Request_attempt_result{.curl_result = res, .response_code = 0};
+        }
+        return Request_attempt_result{.curl_result = CURLE_OK,
+                                      .response_code = Get_response_code(curl)};
```
burgholzer: Feels like this could be simplified further by tweaking `Get_response_code` and removing the conditional here.

src/internal/curl_http_client.cpp
```diff
@@ -390,6 +465,33 @@ void Enable_curl_easy_init_failure_for_testing() {
...
+        }
+        const auto response_code = response_codes[next_response_code_index];
+        next_response_code_index++;
+        return Request_attempt_result{.curl_result = CURLE_OK,
+                                      .response_code = response_code};
+      },
+      [&]([[maybe_unused]] const uint8_t seconds) { sleep_call_count++; });
+  return Retry_test_result{.status_code = status_code,
+                           .sleep_call_count = sleep_call_count};
+}
```
burgholzer: What is test code doing in the src file? This feels pretty wrong.

burgholzer (changes requested):
Thanks, we are getting there.
I believe we shouldn't just let this lie around. Even if we may get a better token eventually, this is a real problem that may occur for users and it should be fixed if possible.

My biggest concern with this PR is the awkward split into source and test code that feels kind of wrong to me.
The actual implementation of the retry is mostly fine for me now.

src/internal/curl_http_client.cpp
marcelwa: I think that was an actual oversight in the code base. [`CURLOPT_SSL_VERIFYPEER`](https://curl.se/libcurl/c/CURLOPT_SSL_VERIFYPEER.html) and [`CURLOPT_SSL_VERIFYHOST`](https://curl.se/libcurl/c/CURLOPT_SSL_VERIFYHOST.html) were set to `0L` for development, which deactivated these security features. In [be49eb4](https://github.com/iqm-finland/QDMI-on-IQM/pull/52/commits/be49eb447db1573668a04821803481bca43de36b), I have adjusted the docstring to match the code changes.

burgholzer (approved):
This should be ready to go in now.
I pushed a couple of cleanups and a changelog entry.
Feel free to merge once you are also happy with the commits and the CI is green!

PR #53, opened 2026-04-30
:sparkles: Add QSCI and MQT Bench end-user showcases

marcelwa:
This PR adds QSCI and MQT Bench showcases that demonstrate how to use the `IQMBackend` with Qiskit to execute meaningful workflows.

In the CI, both showcase categories use MQT Core's DDSIM QDMI device for performance reasons.


marcelwa:
@burgholzer I would like to hear your thoughts on whether to execute these showcase workflows in the CI. The MQT Bench ones are rather lightweight, but the QSCI one is a heavy hitter. We could turn down the iterations and shots, though.

burgholzer:
> @burgholzer I would like to hear your thoughts on whether to execute these showcase workflows in the CI. The MQT Bench ones are rather lightweight, but the QSCI one is a heavy hitter. We could turn down the iterations and shots, though.

None of these benchmarks really make sense on mock-backends as the results are arbitrary.
This is also the main cause for QSCI taking literally forever; it is simply not converging.

I think it would still be helpful to exercise the tests in some kind of way. To ensure that the benchmarks themselves are doing the right thing.
I would argue that the reasonable thing to do is to run the tests through a classical simulator.
Since the qiskit plugin also directly ships mqt-core, we have to the DD-based classical simulator. At the top of my head, I am not 100% about the easiest way to call it though.
Potentially by accessing the respective QDMI device which is part of MQT Core and distributed with the Python package. This might work out of the box already, or may need some additional helpers in MQT Core to get to work. I believe this would make for a nice way to cover the algorithmic functionality added as part of this PR (even though it technically does not really test the IQM QDMI device; but I'd argue that is fine).

marcelwa:
The QSCI showcase is non-deterministic. It just failed on Ubuntu with a violated energy assertion, but passed on a rerun. I see if I can provide a deterministic starting point.

marcelwa:
@burgholzer this is now ready for an initial review. It's okay to stay high-level in the first iteration to align our expectations on the scope and the direction these showcases are taking. Many thanks! :pray:

pyproject.toml
```diff
@@ -72,6 +72,14 @@ test = [
   "pytest-xdist>=3.8.0",
   {include-group = "qiskit"},
 ]
+showcase = [
```
burgholzer: I am pretty sure this should not be a dependency group but rather an optional dependency. Because this is something that users want to specify in their install of the iqm-qdmi package as an extra.
Dependency groups are only available when you have the source checked out.
I would assume that people would want to execute these showcases after they have run uv pip install iqm-qdmi[showcases] or similar. Which also makes me question whether the respective code should be part of the test tree as opposed to the package tree.
On the other hand, one could also argue that it is fine if people need to check out the source code here to run the end2end demos. At that point, it can stay a dependency group.

Edit: now that I thought more about this (this, in combination with the point below, is really bugging me), I think the most pragmatic way to distribute these showcases is as scripts with PEP 723 inline script metadata. This makes them standalone demonstrations where people can simply execute the script to execute the showcase.
This also solves the general dependency problem because it isolates the dependencies to the script itself. This should allow us to use the latest dependencies possible everywhere. We can even limit the applicable Python versions and platforms using `uv` specific metadata in the script blocks (for the QSCI use case).
An open question for me is whether we still want to go with the pytest-style checks or want to replace them with more regular scripting logic.

pyproject.toml
```diff
@@ -72,6 +72,14 @@ test = [
...
+  {include-group = "test"},
+  "mqt-bench>=2.0.0",
+  "pyscf>=2.11.0; platform_system != 'Windows'",
+  "qiskit-algorithms>=0.3.0,<0.4.0; python_version < '3.14'",
+  "qiskit>=1.1,<2.0; python_version < '3.14'",
+  "qiskit-nature>=0.7.2; python_version < '3.14'",
```
burgholzer: This sequence of dependency restrictions is unfortunate as it artificially limits the overall compatibility with qiskit. One can see this in the lockfile, where 1.4.5 is used everywhere except Python 3.14 (where I assume these dependencies are not available).

I know it is not particularly pretty, but might it be feasible to collect the bits and pieces necessary to run QSCI from these dependencies and simply include them as part of the showcase in a Qiskit-2.0 compatible way?
Given how it's all Apache 2.0, there is at least no license obstacle.

MQT Bench dropped support for Qiskit 1.x in v2.2.0 and I really do not feel too comfortable not supporting Qiskit v2.x in the showcases now that 2.x has been around for more than a year.

Edit: Given the edit above, this is not top priority, but I think it would still be nice to have a modern demonstration that does not rely on outdated software packages at some point.

pyproject.toml
```diff
@@ -125,9 +133,20 @@ addopts = [
...
 ]
+markers = [
+  "mqt_bench: marks MQT Bench showcases",
+  "qsci: marks QSCI showcases",
+]
 filterwarnings = [
   "error",
-  "ignore:.*Device operation 'prx_12' cannot be mapped to a Qiskit gate and will be skipped.*:UserWarning:",
+  'ignore:.*Device operation.*cannot be mapped to a Qiskit gate.*:UserWarning',
+  'ignore:Device does not define a measurement operation.*:UserWarning',
```
burgholzer: Updating to `mqt-core~=3.6.0` should make this obsolete.

pyproject.toml
```diff
@@ -125,9 +133,20 @@ addopts = [
...
+  'ignore:qiskit\.providers\.models is deprecated since Qiskit 1\.2.*:DeprecationWarning',
+  'ignore:The class ``qiskit\.circuit\.library\.n_local\.n_local\.NLocal`` is pending deprecation.*:PendingDeprecationWarning',
+  'ignore:The class ``qiskit\.primitives\.backend_estimator\.BackendEstimator`` is deprecated.*:DeprecationWarning',
+  'ignore:The class ``qiskit\.primitives\.base\.base_estimator\.BaseEstimator`` is deprecated.*:DeprecationWarning',
+  'ignore:scipy\.optimize.*L-BFGS-B solver are deprecated.*:DeprecationWarning',
+  'ignore:The function ``qiskit\.primitives\.utils\.init_observable\(\)`` is deprecated.*:DeprecationWarning',
```
burgholzer: I'd hope the proposed change above also makes all of this obsolete.

README.md
```diff
@@ -159,6 +159,28 @@ $ export RESONANCE_API_KEY="your-api-key"
...
 ```
 
+```console
+# Run the optional Python showcases on an IQM backend (default)
+$ export IQM_BASE_URL="https://desired-iqm-server.com"
+$ export RESONANCE_API_KEY="your-api-key"
+$ uv run --group showcase pytest test/showcases
+
+# Run the same showcases against DDSIM instead
+$ IQM_SHOWCASE_BACKEND=ddsim uv run --group showcase pytest test/showcases
```
burgholzer: I think it would be best if we would not explicitly name DDSIM without ever providing context, what it is. I'd rather prefer to simply call this `sim` (or `simulator`) instead of `ddsim`. 
In the end, it does not really matter what kind of simulator is being used.

docs/showcases.md
```diff
@@ -0,0 +1,215 @@
...
+This page showcases end-to-end Qiskit demonstrations built on top of the packaged {py:class}`~iqm.qdmi.qiskit.IQMBackend` and the generic QDMI backend surface exposed by MQT Core.
+Unlike the C API examples in the general [usage guide](usage.md), these demonstrations focus on how end users can apply the `qiskit` extra together with MQT Core's generic sampler and estimator primitives.
+
+The corresponding implementations live in the `test/showcases/` tree:
+
+- `test/showcases/mqt_bench/test_mqt_bench.py`
+- `test/showcases/qsci/test_qsci.py`
+
+These showcases are intentionally kept out of the default Python test path.
+They are heavier than the wrapper tests in `test/python/` and are meant to be run explicitly when you want a realistic end-user demonstration of the QDMI-backed Qiskit integration.
```
burgholzer: For my taste, this is phrased too little like "How can I use this library to perform cool stuff on IQM systems?"
The writing feels fairly technical to me and I think it should be more engaging.

docs/showcases.md
```diff
@@ -0,0 +1,215 @@
...
+:::{important}
+The IQM-backed showcase assertions are tuned for real IQM QPUs.
+If you point `IQM_QC_ALIAS` or `IQM_QC_ID` at a mock IQM target, the showcase is still executed, but the stricter showcase assertions may fail.
+Use `IQM_SHOWCASE_BACKEND=ddsim` if you want a validation path without IQM credentials.
+:::
+
+## MQT Bench Showcase
+
+[MQT Bench](https://mqt.readthedocs.io/projects/bench/) provides a large catalog of benchmark circuits.
+In this repository, it serves as a sampler showcase: the test suite generates benchmark circuits, transpiles them for the selected showcase backend, executes them through {py:class}`~mqt.core.plugins.qiskit.sampler.QDMISampler`, and validates the observed bitstring distribution.
```
burgholzer: I'd probably order these the other way round. QSCI first, MQT Bench second. QSCI is the more impactful demonstration.

docs/showcases.md
```diff
@@ -0,0 +1,215 @@
...
+)
+if not any(instruction.operation.name == "measure" for instruction in circuit.data):
+    circuit.measure_all()
+
+if backend_kind == "ddsim":
+    transpiled = transpile(
+        circuit, basis_gates=sorted(backend.operation_names), optimization_level=1
+    )
+else:
+    transpiled = transpile(circuit, backend=backend, optimization_level=3)
```
burgholzer: Why differentiate on the kind of backend here?
```
transpiled = transpile(circuit, backend=backend, optimization_level=3)
```
should be sufficient in both cases.

docs/showcases.md
```diff
@@ -0,0 +1,215 @@
...
+
+sampler = QDMISampler(backend, default_shots=1024)
+job = sampler.run([(transpiled,)], shots=1024)
```
burgholzer: shots are set twice here. at least one of these is redundant.

docs/showcases.md
```diff
@@ -0,0 +1,215 @@
...
+sample_circuit.measure_all()
+if backend_kind == "ddsim":
+    transpiled = transpile(
+        sample_circuit,
+        basis_gates=sorted(backend.operation_names),
+        optimization_level=1,
+    )
+else:
+    transpiled = transpile(sample_circuit, backend=backend, optimization_level=3)
+```
```
burgholzer: It feels fairly brittle to copy-paste the code here in the docs as it will eventually become outdated. I believe that myst/sphinx has a setting to directly include files that would seem like a reasonable thing to do here.
I'd probably list the description plus options first and then the source code listing.

docs/showcases.md
```diff
@@ -0,0 +1,215 @@
...
+- `IQM_QSCI_ENERGY_TOLERANCE`: accepted deviation from the exact reference energy.
+
+The defaults are tuned for showcase runs, not for aggressive convergence.
+
+## Relationship to the Thin Python Tests
+
+The lightweight tests in `test/python/test_qiskit_backend.py` remain the fast test layer for the backend itself.
+They verify creation plus direct {py:meth}`~mqt.core.plugins.qiskit.backend.QDMIBackend.run`, {py:meth}`~iqm.qdmi.qiskit.IQMBackend.sampler`, and {py:meth}`~iqm.qdmi.qiskit.IQMBackend.estimator` calls.
+
+The showcases in `test/showcases/` build on top of that layer and are intended to answer a different question: what do realistic application-level showcases look like when they utilize the backend?
```
burgholzer: This feels like AI slop to me.

test/showcases/conftest.py
burgholzer, line 1: If we split the use cases up into individual scripts, there will be no place for this module. I am a big fan of self contained scripts. If there is functionality that is needed in multiple places, we may consider adding it to the Python package itself. Otherwise, I believe the code duplication is justified and acceptable.

test/showcases/support.py
```diff
@@ -0,0 +1,206 @@
...
+        circuit: The quantum circuit to transpile.
+
+    Returns:
+        The transpiled circuit ready for execution on the backend.
+    """
+    if showcase_backend_kind() is ShowcaseBackend.DDSIM:
+        # The DDSIM backend's Target metadata does not round-trip through
+        # qiskit.transpile(backend=...), so decompose against the exposed basis
+        # gates without asking Qiskit to build backend properties from Target.
+        return transpile(circuit, basis_gates=sorted(backend.operation_names), optimization_level=1)
```
burgholzer: I think I fixed some of that in the 3.6.0 release of MQT Core.
The main problem here is global gates with properties. This is only properly supported for Qiskit v2.x.
If the use cases are split up into separate scripts, the MQT Bench ones will not need this distinction.
The QSCI example might.

test/showcases/support.py
```diff
@@ -0,0 +1,206 @@
...
+    """Return an estimator primitive for the selected QDMI backend.
+
+    Args:
+        backend: The QDMI backend that should execute the estimator jobs.
+        shots: The default shot budget for expectation-value estimation.
+
+    Returns:
+        A :class:`~mqt.core.plugins.qiskit.estimator.QDMIEstimator` bound to the selected backend.
+    """
+    return QDMIEstimator(backend, options={"default_shots": shots})
```
burgholzer: These are wrappers around a one-liner. I do not believe they provide real value and rather view them as bloat.

test/showcases/support.py
```diff
@@ -0,0 +1,206 @@
...
+
+    Returns:
+        A circuit that includes measurements on all qubits.
+    """
+    measured_circuit = circuit.copy()
+    if any(instruction.operation.name == "measure" for instruction in measured_circuit.data):
+        return measured_circuit
+
+    measured_circuit.measure_all()
+    return measured_circuit
```
burgholzer: Why the copy? feels unnecessary and harms performance unnecessarily

test/showcases/support.py
```diff
@@ -0,0 +1,206 @@
...
+    data = job.result()[0].data
+    if "meas" in data:
+        return data["meas"].get_counts()
+
+    register_name = next(iter(data.keys()), None)
+    if register_name is None:
+        msg = "Sampler result did not contain any classical measurement data."
+        raise ValueError(msg)
+
+    return data[register_name].get_counts()
```
burgholzer: I feel like this also becomes way cleaner when integrated into dedicated scripts because there is no guesswork necessary for how to interpret/get the results -> less bloat.

test/showcases/mqt_bench/test_mqt_bench.py
burgholzer, line 1: When splitting up, it might make sense to simply make these 5 scripts.
That makes each one self-contained and pretty clear.

test/showcases/qsci/test_qsci.py
```diff
@@ -0,0 +1,209 @@
...
+    "qiskit_nature.second_q.algorithms.initial_points", reason=QSCI_PY314_REASON
+)
+qiskit_nature_mappers = pytest.importorskip("qiskit_nature.second_q.mappers", reason=QSCI_PY314_REASON)
+pytest.importorskip(
+    "pyscf",
+    reason=QSCI_WINDOWS_REASON,
+)
+
+postprocess = importlib.import_module("postprocess")
+support = importlib.import_module("support")
```
burgholzer: All of this becomes much cleaner when this is a simple script.

burgholzer (changes requested):
Thanks @marcelwa for pushing this!
I now had a detailed look through the implementation here and left a couple of comments with feedback.
I worked my way from the configuration over the documentation to the actual implementation (maybe one can see this in the actual comments).

My main concern is how this is generally set up and distributed. However, I think I found a pretty neat solution that I tried to outline in the comments. I hope this makes sense.
On a secondary note, I think the phrasing of parts of the docs should be more "human-like"/"human-friendly", which I am currently missing a bit when reading through.

marcelwa:
@burgholzer many thanks for the suggestion of refactoring the examples into individual self-contained scripts. That was a great idea and simplified so many things! :pray: The error we're seeing here on Windows is the same one as in #63. I would much appreciate it if you could have another look at the changes.

examples/qsci_h2.py
```diff
@@ -0,0 +1,328 @@
...
+from qiskit_nature.second_q.circuit.library import UCCSD, HartreeFock
+from qiskit_nature.second_q.drivers import PySCFDriver
+from qiskit_nature.second_q.mappers import JordanWignerMapper
+from scipy.sparse import csr_matrix
+from scipy.sparse.linalg import eigs
+
+from iqm.qdmi.qiskit import IQMBackend
+
+if TYPE_CHECKING:
+    from qiskit.quantum_info import SparsePauliOp
```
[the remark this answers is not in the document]
burgholzer: This is a false positive. The import is used as part of the `cast` expression further down the file.

burgholzer (approved):
Thanks @marcelwa for the initial work on this.
I pushed quite some updates to this PR now that streamline a lot of the implementation here as well as the overall documentation to make things more approachable.
It's probably best if you go through the individual commits.

CI seems to be happy so far. And I am too. If you are too, then let's get this in.
I'll set up a PR that prepares the next minor release directly after that.

docs/spack_guide.md
```diff
@@ -1,47 +1,111 @@
...
 
-    depends_on("cmake", type="build")
-    depends_on("curl", type=("build", "link", "run"))
-    depends_on("nlohmann-json")
+       maintainers("@burgholzer", "@marcelwa")
 
-    def cmake_args(self):
-        args = [self.define("BUILD_IQM_QDMI_TESTS", False)]
-        return args
+       version("1.0.1", tag="v1.0.1", commit="5d37eca717051f7a32b70762f3e0d8d8f457d36c")
```
marcelwa: Concrete tags (`1.0.1` in this case) have been added throughout this file and would need to be updated for every release. That is a cumbersome chore, easy to forget, and error-prone when done manually. I'd prefer a placeholder rather than a specific tag.

noxfile.py
```diff
@@ -115,6 +116,32 @@ def minimums(session: nox.Session) -> None:
...
+
+    env = {"UV_PROJECT_ENVIRONMENT": session.virtualenv.location}
+    for script in Path("examples").glob("*.py"):
+        session.run(
+            script,
+            "--backend",
+            args.backend,
+            env=env,
+            external=True,
+        )
```
marcelwa, line 142: I like the simplification here, but grabbing all Python files in `examples/` means we can't have helper classes/functions in separate files. Like how we had the dedicated `postprocess.py` for the QSCI workflow. Not a big issue since you merged it into the `qsci_h2.py` anyway. And I don't think it's a sufficiently severe future-proofing consideration to mandate reverting your changes. Just a thought.

marcelwa (commented):
Many thanks for your edits, @burgholzer 🙏 I carefully went through your commits and agree with almost everything. Two very minor considerations came to mind that you can find below. We can probably consolidate here super fast and get the PR merged.

docs/spack_guide.md
burgholzer: Hm, my thinking was that I'd rather have something that people can literally copy&paste instead of having to adapt and leaving some guesswork on how exactly to add something.
I am not hard set about this. Feel free to push an update if you have a better suggestion here.

noxfile.py
burgholzer: IMO the examples should be self-contained. That's the entire point of them being scripts.

The examples directory is not a package (has no `__init__.py`) and should not become one. As such, importing between different files is somewhat broken anyway.

Overall I do not see a problem with this now and also not in the short term. We can always pivot in the future if we wanted to.

docs/spack_guide.md
marcelwa: If we could fetch the latest tag at build time and insert it into this page, that would solve both issues. I'll check if there is a simple solution for this.

noxfile.py
marcelwa: Makes sense to me. Let's keep it as is.

docs/spack_guide.md
burgholzer: Maybe https://github.com/munich-quantum-toolkit/core/blob/main/cmake/GetVersion.cmake helps for inspiration. That's how we handle the versioning of the MQT Core CMake project.

marcelwa: I played around with [sphinx-substitution-extensions](https://github.com/adamtheturtle/sphinx-substitution-extensions) a bit as it seemed like a straightforward solution to our problem. However, integrating it with MyST and executable notebooks requires quite some code overhead. So I opted for a simpler solution and wrote a small Python Sphinx extension that does the same thing. It extracts git tags, release versions, and commits from the git history and injects them into docs files by replacing placeholders. Let me know what you think.

docs/_ext/release_substitutions.py
```diff
@@ -0,0 +1,115 @@
...
+class _SphinxConfig(Protocol):
+    release: str
+    release_substitutions: dict[str, str]
+
+
+class _SphinxApp(Protocol):
+    confdir: str
+    config: _SphinxConfig
+
+    def add_config_value(self, name: str, default: dict[str, str], rebuild: str) -> None: ...
```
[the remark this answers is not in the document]
marcelwa: This is a protocol method, so this finding is a false positive.

docs/_ext/release_substitutions.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def connect(self, event: str, callback: Callable[..., object]) -> None: ...
```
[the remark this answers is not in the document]
marcelwa: This is a protocol method, so this finding is a false positive.

docs/spack_guide.md
burgholzer: Feels a bit overkill to me to be honest. I just checked, and why can't we simply use MyST's subsitution feature for this? https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#substitutions-with-jinja2

The version is already available in `conf.py` (as `release`) and the commit SHA can be derived from in the conf file as well supposedly.
Feels like a 5-10 line change overall, compared to 100+ lines of extension code.

marcelwa: I get that. I tried to make it work with both the substitutions extension mentioned above and MyST's built-in feature. However, it does not work out of the box with fenced Python code blocks or executable inline notebooks. MyST doesn't treat it as raw Markdown, so substitutions won't take effect.

burgholzer: Too bad.. honestly speaking I do not think that the added complexity is worth the benefit. At this point, I'd rather list no explicit version and just use a placeholder. Doesn't require maintenance from our side and doesn't require additional workarounds at the downside of people having to fill in the blanks.
Would you agree?

marcelwa: Absolutely. I'll revert the changes.

burgholzer (approved):
LGTM. Let's get this in 🚀

PR #56, opened 2026-05-04
🚚 Replace meetiqm.com with the new domain

iqmtjm:
IQM is changing the domain name from `meetiqm.com` to `iqm.tech` so the code and documents are updated accordingly

The update is starting 4.5.2026 and will rolled out over time so some domains might still re-direct to the old meetiqm.com for some time


burgholzer (approved):
LGTM 👍🏼

PR #67, opened 2026-05-21
:recycle: Consistently prefer `IQM_TOKEN` instead of `RESONANCE_API_KEY` for authentication

marcelwa:
Most of the code base used `IQM_TOKEN` and `IQM_TOKENS_FILE` for authentication. However, the `IQMBackend` (and its tests) and the C++ integration test suite would still use `RESONANCE_API_KEY`. The documentation was also inconsistently using either. 

This PR cleans up this oversight so that `IQM_TOKEN` and `IQM_TOKENS_FILE` are consistently used. `RESONANCE_API_KEY` will no longer be supported anywhere.

Note that the respective secret here on GitHub is still called `RESONANCE_API_KEY`, but it is now loaded into an environment variable called `IQM_TOKEN`.

These changes revealed another subtle inconsistency between the C++ and Python authentication contracts: in Python, explicit authentication parameters take precedence over environment variables, as the docs also state. In C++, however, setting both would trigger an exception. This PR hence further streamlines the contract by aligning the C++ implementation and test suite with the Python implementation and docs. This is an important change for the upcoming SPANK plugin as well.

Furthermore, the C++ tests could leak secrets to the console in certain cases. This has been fixed as well.


burgholzer:
This looks great from a quick glance.
Note that this is a breaking change and definitely deserves a changelog entry.

burgholzer (approved):
LGTM. I assume all the failures in CI are due to hitting rate limits.
Let's just retry them a couple of times until we are green here.
I suppose we can prepare the 1.1.0 release afterwards.

PR #74, opened 2026-05-22
:sparkles: Add IQM SPANK plugin

marcelwa:
This PR adds an IQM SPANK plugin that facilitaces integration with HPC setups using Slurm. It enables environment injection through admins (pre-defined defaults) and users (custom values) for authentication, endpoints, etc.


spank/.license-tools-config.json
```diff
@@ -0,0 +1,41 @@
+{
+  "author": {
+    "name": "IQM Finland Oy",
+    "years": [2025, 2026]
+  },
+  "force_author": true,
+  "custom_license": "This program is free software: you can redistribute it and/or modify it under\nthe terms of the GNU General Public License as published by the Free Software\nFoundation, version 3.\n\nThis program is distributed in the hope that it will be useful, but WITHOUT\nANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS\nFOR A PARTICULAR PURPOSE. See the GNU General Public License for more\ndetails.\n\nYou may obtain a copy of the License at\n\nhttps://github.com/iqm-finland/QDMI-on-IQM/blob/main/spank/LICENSE.md\n\nSPDX-License-Identifier: GPL-3.0-only",
```
burgholzer: Does this need to be a custom license? GPLv3 is a standard license after all.

spank/.license-tools-config.json
```diff
@@ -0,0 +1,41 @@
+{
+  "author": {
+    "name": "IQM Finland Oy",
+    "years": [2025, 2026]
```
burgholzer: Shouldn't this just be 2026?

README.md
```diff
@@ -57,5 +58,10 @@ the development workflow, coding standards, and pull request process.
 
 ## License
 
-This project is licensed under the Apache License 2.0 with LLVM exception.
-See the [LICENSE.md](LICENSE.md) file for details.
+The code in this repository is available under two licenses:
```
burgholzer: This sounds a bit like it could be misinterpreted as dual licensing, which is not what this is.
Can we make this a bit more explicit/less ambiguous?

spank/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+# https://github.com/iqm-finland/QDMI-on-IQM/blob/main/spank/LICENSE.md
+#
+# SPDX-License-Identifier: GPL-3.0-only
+
+set(BUILD_IQM_QDMI_SPANK_TARGET iqm-qdmi-spank-plugin)
+add_library(${BUILD_IQM_QDMI_SPANK_TARGET} MODULE iqm_qdmi_spank_plugin.cpp)
+target_compile_features(${BUILD_IQM_QDMI_SPANK_TARGET} PRIVATE cxx_std_20)
+target_include_directories(${BUILD_IQM_QDMI_SPANK_TARGET}
+                           PRIVATE ${SLURM_SPANK_INCLUDE_DIR})
+set_target_properties(${BUILD_IQM_QDMI_SPANK_TARGET} PROPERTIES PREFIX "")
```
burgholzer: What is the purpose of this line here?

spank/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+#
+# https://github.com/iqm-finland/QDMI-on-IQM/blob/main/spank/LICENSE.md
+#
+# SPDX-License-Identifier: GPL-3.0-only
+
+set(BUILD_IQM_QDMI_SPANK_TARGET iqm-qdmi-spank-plugin)
+add_library(${BUILD_IQM_QDMI_SPANK_TARGET} MODULE iqm_qdmi_spank_plugin.cpp)
+target_compile_features(${BUILD_IQM_QDMI_SPANK_TARGET} PRIVATE cxx_std_20)
+target_include_directories(${BUILD_IQM_QDMI_SPANK_TARGET}
+                           PRIVATE ${SLURM_SPANK_INCLUDE_DIR})
```
burgholzer: This is very likely me being stupid, but shouldn't the spank plugin link against the `iqm-qdmi` library? Or is the idea, that the QDMI device library will always be provided via the Python package distribution?

spank/README.md
```diff
@@ -0,0 +1,72 @@
+# IQM QDMI SPANK Plugin
+
+[![SPANK Plugin GPLv3 License](https://img.shields.io/static/v1?logo=gnu&label=License&message=GPLv3&color=informational&style=flat-square)](https://www.gnu.org/licenses/gpl-3.0.en.html)
+[![C++20](https://img.shields.io/static/v1?logo=cplusplus&label=C%2B%2B&message=20&color=informational&style=flat-square)](https://isocpp.org/)
+[![Slurm SPANK](https://img.shields.io/static/v1?label=Slurm&message=SPANK&color=informational&style=flat-square)](https://slurm.schedmd.com/spank.html)
+
+**IQM QDMI SPANK Plugin** integrates the QDMI-on-IQM runtime with Slurm through
+the [SPANK plugin interface](https://slurm.schedmd.com/spank.html).
+It injects the canonical `IQM_*` environment contract into Slurm jobs, registers
```
burgholzer: "environment contract" screams AI to me.

spank/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+Inside your script, the `IQMBackend` (from `iqm-qdmi`) will automatically pick up these settings—no manual `export` or hardcoded URLs required.
+
+## Where to Start
+
+| I want to…                                                | Guide                                         |
+| :-------------------------------------------------------- | :-------------------------------------------- |
+| Understand the plugin runtime contract and user workflows | [SPANK Plugin Guide](../docs/spank_plugin.md) |
+| Configure plugstack entries and site defaults             | [SPANK Plugin Guide](../docs/spank_plugin.md) |
+| Build the plugin together with QDMI-on-IQM                | [Root README](../README.md)                   |
```
burgholzer: None of these point to the rest of the README. This feels odd.

spank/README.md
burgholzer, line 1: We should probably mention some kind of compatibility with SLURM versions (older is better).

spank/README.md
```diff
@@ -0,0 +1,72 @@
...
+portable binaries.
+
+Build against the target cluster's Slurm headers and rebuild after Slurm
+major-version upgrades.
+
+## License and Compatibility
+
+The SPANK plugin in this directory is licensed under GPLv3 because it links against
+Slurm's GPL-licensed SPANK interface. See [spank/LICENSE.md](LICENSE.md) for the
+plugin license text.
```
burgholzer: I am personally not the biggest fan of the duplication between the separate README and the main documentation. I'd prefer to simply link to the main documentation after a fairly minimal introduction here + the license notice.

docs/spank_plugin.md
```diff
@@ -0,0 +1,222 @@
...
+code.
+
+If you use `IQM_TOKENS_FILE`, the path must exist and be readable on the compute
+nodes where tasks run, not only on the submit node.
+
+### Practical Examples
+
+#### Using Python and Qiskit
+
+Most users will interact with our hardware via Python. The `IQMBackend` automatically picks up the settings injected by the plugin. Here is a complete "Hello World" example that creates a Bell state:
```
burgholzer: Proper xref for the Backend?

docs/spank_plugin.md
```diff
@@ -0,0 +1,222 @@
...
+
+**To run it:**
+
+```bash
+srun --partition=quantum --iqm-qc-alias=emerald python bell_state.py
+```
+
+#### Using Native C++ Applications
+
+If you are developing low-level applications, you can use the QDMI C++ library. The plugin ensures that `IQM_QDMI_device_session_init` finds everything it needs.
```
burgholzer: I think this section is mostly fine for now; but it is more so showing how to use the device from C using the very low-level functions.
For practical purposes, people would probably rely on the C++-based FoMac we offer through MQT Core and the general QDMI Driver instead of the specific IQM QDMI device.
But, as said, just fine for now.

docs/spank_plugin.md
```diff
@@ -0,0 +1,222 @@
...
+# Compile using paths from the iqm-qdmi utility
+g++ -std=c++20 device_info.cpp -o device_info \
+  -I"$(iqm-qdmi --include_dir)" \
+  -L"$(dirname "$(iqm-qdmi --lib_path)")" \
+  -Wl,-rpath,"$(dirname "$(iqm-qdmi --lib_path)")" \
+  -liqm_qdmi_device
+
+# Run it on the cluster
+srun --partition=quantum ./device_info
+```
```
burgholzer: Stupid me, but can this actually work as described? This builds on the login node, but then executes on a compute node. Wouldn't this mean that the built executable (as well as the shared IQM QDMI device library) need to be available on the compute node as well?
I am probably overlooking something and you have thought this through, but maybe you could help me out here.

spank/iqm_spank_plugin.cpp
```diff
@@ -0,0 +1,647 @@
...
+  slurm_spank_log("[iqm_spank_plugin] hook=%s", hook_name);
+}
+
+} // namespace
+
+// NOLINTBEGIN(readability-identifier-naming)
+
+extern "C" {
+/// Spank plugin name. Must be unique across all loaded plugins.
+extern const char plugin_name[] = "iqm_spank_plugin";
```
burgholzer, line 561: should this contain QDMI as well? probably not, right?

spank/iqm_spank_plugin.cpp
```diff
@@ -0,0 +1,647 @@
...
+
+/**
+ * @brief SPANK post-option-init hook.
+ * @return `ESPANK_SUCCESS`.
+ */
+int slurm_spank_init_post_opt(spank_t /* spank */, const int /* ac */,
+                              char ** /* av */) {
+  Emit_hook_log("slurm_spank_init_post_opt");
+  return ESPANK_SUCCESS;
+}
```
burgholzer, line 592: Just for understanding: this has to be implemented even though it does nothing, right?

burgholzer (changes requested):
Hey @marcelwa 👋🏼 
I know that you are still ironing out a few kinks, but I figured I'd give this an initial review to keep the great momentum rolling here.
This look like great work 🙌🏼 Almost none of my comments are on the SPANK plugin itself, which looks very reasonable. Most are on wording and docs, which should be fairly quick to fix.
There are some questions for my understanding. I hope they are not too stupid. Pardon me, if they are.

Keep it up 🤝

spank/.license-tools-config.json
marcelwa: I guess that's right. We used 2025--2026 for the QDMI device. However, since the SPANK plugin is its own entity with its own license, adding 2026 makes sense.

spank/CMakeLists.txt
marcelwa: Currently, that is the idea, yes.

spank/README.md
marcelwa: Yup, I wasn't done streamlining this :sweat_smile:

spank/CMakeLists.txt
marcelwa: It is to prevent CMake from prepending `lib` to the SPANK plugin binary, as that would violate the Slurm naming conventions. We want it to be called `iqm-qdmi-spank-plugin.so`, not `libiqm-qdmi-spank-plugin.so`.

burgholzer: AH. Thanks for the clarification. That makes a lot of sense!

spank/iqm_spank_plugin.cpp
marcelwa: I was debating that myself, and am still unsure. As evident by the filename, I went through several cycles here :joy: There are pros and cons to both. Do you have any preference?

spank/iqm_spank_plugin.cpp
marcelwa: Yes, the function has to exist; otherwise, we risk shooting ourselves in the foot.

spank/iqm_spank_plugin.cpp
burgholzer: Hm. Technically, in its current form, the spank plugin can also be used with the iqm-client (which reads the same env vars). So staying a little more generic is probably the slightly better option.

docs/spank_plugin.md
marcelwa: Good point. I was assuming a shared filesystem between the login and compute nodes. If that is not given, this piece cannot work, of course. I added a note to clarify that.

marcelwa: We could resort to static linking, I guess...

spank/iqm_spank_plugin.cpp
marcelwa: That's a great argument. I would then rename the file to `iqm_spank_plugin.cpp` as well and adjust the documentation accordingly.

docs/spank_plugin.md
burgholzer: Even for static linking, the executable would have to be moved to the compute node.

Making the assumption explicit that this relies on a shared filesystem partition between the nodes is certainly a good idea.
Another idea would be to remove this example all together and add it back, once a real deployment scenario has presented itself.

spank/iqm_spank_plugin.cpp
burgholzer: Works for me

docs/spank_plugin.md
marcelwa: That would certainly be the safest bet. I think I prefer that solution for the time being.

.github/workflows/CI.yml
```diff
@@ -44,6 +44,14 @@ jobs:
     secrets:
       IQM_TOKEN: ${{ secrets.RESONANCE_API_KEY }}
 
+  spank-tests:
+    name: 🇨 Test SPANK
+    needs: change-detection
+    if: fromJSON(needs.change-detection.outputs.run-cpp-tests)
```
marcelwa: We currently run SPANK tests on the `run-cpp-tests` trigger. It's not entirely correct, but I guess we shouldn't mess with the reusable workflows just for this.

burgholzer: The workflows just recently got some more flexibility for change detection: https://github.com/munich-quantum-toolkit/workflows/blob/main/UPGRADING.md#203
maybe some of that proves useful.

marcelwa:
@burgholzer the SPANK plugin is nearing completion. I would love your opinion on the current state of work. Don't mind the failing tests. There is some maintainance going on at Resonance.

README.md
```diff
@@ -57,5 +58,7 @@ the development workflow, coding standards, and pull request process.
 
 ## License
 
-This project is licensed under the Apache License 2.0 with LLVM exception.
-See the [LICENSE.md](LICENSE.md) file for details.
+This repository is licensed under two separate licenses, depending on the component:
```
burgholzer:
```suggestion
The components of this repository are licensed under two separate licenses:
```

spank/.license-tools-config.json
```diff
@@ -0,0 +1,41 @@
+{
+  "author": {
+    "name": "IQM Finland Oy",
+    "years": [2026]
+  },
+  "force_author": true,
+  "license": "GPL-3.0-only",
```
burgholzer:
```suggestion
  "license": "GPL-3.0-or-later",
```
I believe this is also what Slurm is now licensed under and feels slightly more correct.

.github/workflows/cpp-linter.yml
```diff
@@ -53,7 +53,7 @@ jobs:
       - name: Install Dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install -y libcurl4-openssl-dev
+          sudo apt-get install -y libcurl4-openssl-dev libslurm-dev
```
burgholzer, line 56: Will this pull a recent-enough version?

.github/workflows/CI.yml
```diff
@@ -170,6 +170,14 @@ jobs:
     secrets:
       IQM_TOKEN: ${{ secrets.RESONANCE_API_KEY }}
 
+  spank-tests:
+    name: 🇸 Test SPANK
+    needs: change-detection
+    if: fromJSON(needs.change-detection.outputs.run-cpp-tests)
+    uses: ./.github/workflows/spank-tests.yml
+    with:
+      runs-on: ubuntu-24.04
```
burgholzer:
```suggestion
    uses: ./.github/workflows/spank-tests.yml
```
This is the default anyway

.github/workflows/spank-tests.yml
burgholzer, line 1: I generally like this quite a bit!
However, wouldn't it be even nicer if someone could also run the same test locally?
This could, for example, be realized with a Dockerfile that sets everything up.
Not a must for this PR (let's rather get this in), but for a follow-up.

.github/workflows/spank-tests.yml
```diff
@@ -0,0 +1,144 @@
...
+          sudo scontrol reconfigure
+          sleep 2
+
+      - name: Run SPANK smoke tests
+        run: |
+          uv run nox -f spank/noxfile.py -s smoke_tests -- \
+            --partition debug \
+            --hook-mode full \
+            --require-all-hooks \
+            --test-base-url "$TEST_BASE_URL"
```
burgholzer, line 148: Maybe this is just me, but shouldn't this execute a "practical" workload against a Resonance mock like the other CI tests?
Such a run need not replace the smoke tests here, but complement them.

docs/spank_plugin.md
```diff
@@ -0,0 +1,134 @@
+# Slurm SPANK Plugin
+
+The Slurm SPANK plugin for QDMI-on-IQM simplifies running quantum jobs on clusters by automatically propagating `IQM_*` environment variables to job steps. This avoids manual `export` statements in job scripts and enables administrators to configure global defaults and partition-gated access.
```
burgholzer, line 3: Something that came up in some offline discussions: The wording throughout some of the docs makes it seem like the only purpose of the Spank plugin is to propagate environment variables, while the main responsibility of the SLURM plugin is to register a QPU as a resource to SLURM.
At least that is what I would argue is the main advantage here. Maybe the wording in this regard could be updated slightly.

Something I was wondering in that regard: We currently have no way to give the user any kind of indication, which QPUs are available to them, right? Some kind of device/resource discovery.
Probably too much to ask for the initial PR, but likely interesting for a follow-up.

docs/spank_plugin.md
```diff
@@ -0,0 +1,134 @@
...
+- **C++ Compiler**: C++20 standard library support (GCC 13+ or Clang 16+).
+- **Compilation Constraint**: SPANK plugins are tied to the Slurm daemon ABI. You must compile the plugin against the target cluster's Slurm header files (`slurm/spank.h`) and rebuild the plugin after any major/minor Slurm upgrades.
+
+### Installation
+
+To compile and install the plugin from the repository root:
+
+```bash
+cmake -S . -B build-spank -DBUILD_IQM_SPANK=ON
+cmake --build build-spank -j
```
burgholzer: Similar to the CI, this should probably include the `--target` flag. (and the below command should contain the component flag)

spank/test/test_env_injection.sh
```diff
@@ -0,0 +1,340 @@
...
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+# Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program. If not, see <https://www.gnu.org/licenses/>.
+
+# Test script for the IQM QDMI SPANK plugin environment injection.
```
burgholzer:
```suggestion
# Test script for the IQM SPANK plugin environment injection.
```

spank/test/test_env_injection.sh
```diff
@@ -0,0 +1,340 @@
...
+#
+# Requires:
+#   - A running Slurm cluster with an allocatable node.
+#   - The IQM QDMI SPANK plugin installed and referenced in plugstack.conf
```
burgholzer:
```suggestion
#   - The IQM SPANK plugin installed and referenced in plugstack.conf
```

spank/test/test_env_injection.sh
```diff
@@ -0,0 +1,340 @@
...
+
+# ---------------------------------------------------------------------------
+# Usage
+# ---------------------------------------------------------------------------
+
+usage() {
+  cat <<'EOF'
+Usage: spank/test/test_env_injection.sh [options]
+
+Tests the IQM QDMI SPANK plugin environment injection behavior.
```
burgholzer:
```suggestion
Tests the IQM SPANK plugin environment injection behavior.
```

spank/test/test_env_injection.sh
```diff
@@ -0,0 +1,340 @@
...
+#   3. Unconfigured variables are absent from the job environment.
+#   4. Conflicting auth is rejected.
+#   5. srun options have highest precedence.
+#   6. Missing IQM_BASE_URL is rejected.
+
+set -euo pipefail
+
+partition=""
+srun_cmd="srun"
+test_base_url="https://test.example.com"
```
burgholzer, line 39: This is used in some other places as well. Why not use the proper resonance domain?

spank/test/test_env_injection.sh
```diff
@@ -0,0 +1,340 @@
...
+if ! command -v "$srun_cmd" >/dev/null 2>&1; then
+  echo "ERROR: '$srun_cmd' not found in PATH." >&2
+  exit 127
+fi
+
+# ---------------------------------------------------------------------------
+# Connectivity check
+# ---------------------------------------------------------------------------
+
+echo "=== IQM QDMI SPANK Plugin: Environment Injection Tests ===" >&2
```
burgholzer:
```suggestion
echo "=== IQM SPANK Plugin: Environment Injection Tests ===" >&2
```

spank/CMakeLists.txt
```diff
@@ -0,0 +1,78 @@
...
+configure_file(iqm-qdmi.conf.in iqm-qdmi.conf @ONLY)
+
+install(TARGETS ${BUILD_IQM_SPANK_TARGET}
+        LIBRARY DESTINATION ${IQM_QDMI_SPANK_INSTALL_DIR}
+                COMPONENT iqm_spank_plugin_Runtime)
+
+install(
+  FILES ${CMAKE_CURRENT_BINARY_DIR}/iqm-qdmi.conf
+  DESTINATION ${IQM_QDMI_SLURM_CONF_DIR}/plugstack.conf.d
+  COMPONENT iqm_spank_plugin_Runtime)
```
burgholzer:
```suggestion
install(TARGETS ${BUILD_IQM_SPANK_TARGET}
        LIBRARY DESTINATION ${IQM_QDMI_SPANK_INSTALL_DIR}
                COMPONENT iqm_spank_plugin)

install(
  FILES ${CMAKE_CURRENT_BINARY_DIR}/iqm-qdmi.conf
  DESTINATION ${IQM_QDMI_SLURM_CONF_DIR}/plugstack.conf.d
  COMPONENT iqm_spank_plugin)
```
Does seem a little cleaner to me as there is no other spank component.

burgholzer (approved):
This generally looks good to me! Thanks for all the work.
I have a couple of nitpick comments left for cleanup and a couple of potential follow ups that should not be blockers here.
Let's try to get this in soon 😌

spank/test/test_env_injection.sh
marcelwa: Good catch. I thought I got them all 😬

.github/workflows/cpp-linter.yml
marcelwa: The version in the Ubuntu 24.04 repos is `23.11.4`. We say we support `17.11.0` or newer.

.github/workflows/spank-tests.yml
marcelwa: Great idea! I'll put this on the list for a follow-up PR.

docs/spank_plugin.md
marcelwa: I have to see how that would work. But I can for sure put it on the list and let you know what I find.

spank/test/test_env_injection.sh
marcelwa: I guess we could use a proper resonance domain here. The point of the tests is to verify whether we can properly set and override it. So it doesn't really matter in the end.

docs/spank_plugin.md
marcelwa: Done.

.github/workflows/spank-tests.yml
marcelwa: I've added a SPANK Resonance workflow in [a7367d5](https://github.com/iqm-finland/QDMI-on-IQM/pull/74/commits/a7367d5b397356b45730decbcecfedb7545815ea).

.github/workflows/CI.yml
marcelwa: Thanks for the pointer. Not quite sure I implemented this correctly. Would need your input on that: https://github.com/iqm-finland/QDMI-on-IQM/pull/74/changes#diff-3ab46ee209a127470fce3c2cf106b1a1dbadbb929a4b5b13656a4bc4ce19c0b8R21-R184

.github/workflows/cpp-linter.yml
burgholzer: Have you checked whether everything also works with the latest SLURM version? (26.x something)
Would be good to at least know that.

marcelwa: I tested locally with `25.11.6` and did not encounter any issues.

burgholzer: Alright. Sufficiently confident that this works then. Feel free to resolve.

burgholzer (approved):
LGTM 👍🏻 let's get this in once CI is green

PR #92, opened 2026-06-18
:sparkles: Add dedicated sampler and estimator CLI scripts

marcelwa:
This PR adds dedicated sampler and estimator convenience CLI scripts that utilize the `IQMBackend.sampler()` and `IQMBackend.estimator()`. 

These additions will ultimately enable a Slurm offloader CLI script that complements #74.


pyproject.toml
```diff
@@ -126,6 +130,16 @@ addopts = [
 ]
 filterwarnings = [
   "error",
+  'ignore:\s.*Pyarrow.*:DeprecationWarning:',
+  'ignore:.*datetime\.datetime\.utcfromtimestamp.*:DeprecationWarning:',
+  # Qiskit 1.2/1.3 deprecations
+  'ignore:.*no need for these schema-conformant objects.*:DeprecationWarning:',
+  'ignore:.*``qiskit.dagcircuit.dagcircuit.DAGCircuit.*`` is deprecated as of qiskit 1.3.0.*:DeprecationWarning:',
+  'ignore:.*``qiskit.circuit.library.standard_gates.x.*`` is pending deprecation as of qiskit 1.3.*:PendingDeprecationWarning:',
+  'ignore:.*``qiskit.circuit.instruction.Instruction.condition`` is deprecated as of qiskit 1.3.0.*:DeprecationWarning:',
```
burgholzer: Why would these still be necessary when the default Qiskit version we resolve for is 2.x?

pyproject.toml
```diff
@@ -126,6 +130,16 @@ addopts = [
...
+  # SciPy solver deprecation
+  'ignore:.*The `disp` and `iprint` options of the L-BFGS-B solver are deprecated.*:DeprecationWarning:',
```
burgholzer, line 136: Is this something we have control over or does this come from Qiskit Algorithms?

docs/python_package.md
```diff
@@ -60,3 +60,38 @@ The above values can also be conveniently queried from the command line via the
...
+If you install the package with the `qiskit` extra, the following additional command-line scripts are exposed:
+
+- `iqm-sampler`: Samples a serialized QPY circuit on the specified backend.
+- `iqm-estimator`: Variational Quantum Eigensolver (VQE) parameter estimation for a serialized ansatz and observable.
+
+### `iqm-sampler` Usage
+
+```console
+$ iqm-sampler --help
+```
```
burgholzer:
````suggestion
```{code-cell} ipython3
!iqm-sampler --help
```
````

File: docs/python_package.md [hunk 4]
```diff
@@ -60,3 +60,38 @@ The above values can also be conveniently queried from the command line via the
...
+
+```console
+$ iqm-sampler bell.qpy --shots 128
+```
+
+### `iqm-estimator` Usage
+
+```console
+$ iqm-estimator --help
+```
```
burgholzer:
````suggestion
```{code-cell} ipython3
!iqm-estimator --help
```
````

File: docs/python_package.md [hunk 5]
```diff
@@ -60,3 +60,38 @@ The above values can also be conveniently queried from the command line via the
...
+
+To run a parameter estimation job:
+
+```console
+$ iqm-estimator ansatz.qpy observable.pkl --maxiter 10
+```
+
+:::{note}
+Direct token passing via a command-line flag is intentionally unsupported to prevent exposing access tokens in shell history, process listings, or logs. To run authenticated jobs, use the `--tokens-file` option or set the `IQM_TOKEN` / `IQM_TOKENS_FILE` environment variables.
+:::
```
burgholzer: Doesn't really feel like this note is warranted here given how the above suggestion will actually print the help text from both entry points.

```suggestion
```

docs/python_package.md
```diff
@@ -60,3 +60,38 @@ The above values can also be conveniently queried from the command line via the
 ```{code-cell} ipython3
 !iqm-qdmi --lib_path
 ```
+
+## Sampler and Estimator CLI Utilities
+
+If you install the package with the `qiskit` extra, the following additional command-line scripts are exposed:
+
+- `iqm-sampler`: Samples a serialized QPY circuit on the specified backend.
+- `iqm-estimator`: Variational Quantum Eigensolver (VQE) parameter estimation for a serialized ansatz and observable.
```
burgholzer: These should probably have cross references to the actual API docs of the entry point modules

docs/spank_plugin.md
```diff
@@ -66,6 +66,18 @@ To run this job:
 srun --partition=quantum --iqm-qc-alias=emerald python bell_state.py
 ```
 
+### Executing via CLI Scripts
+
+Alternatively, if you already have serialized circuits (in QPY format), you can execute them directly on the cluster using the packaged CLI scripts without writing any custom Python code:
```
burgholzer: Some kind of cross-reference to the main documentation of these entry points would be helpful here.

python/iqm/qdmi/estimator.py
```diff
@@ -0,0 +1,112 @@
...
+    from qiskit.quantum_info import SparsePauliOp
+
+
+def build_simulator_backend() -> QDMIBackend:
+    """Create the MQT Core DDSIM backend used for simulation.
+
+    Returns:
+        The MQT Core QDMI simulator backend.
+    """
+    return QDMIProvider().get_backend("MQT Core DDSIM QDMI Device")
```
burgholzer: This is just a one-liner used a single time. I'd simply inline the call.

python/iqm/qdmi/estimator.py
```diff
@@ -0,0 +1,112 @@
...
+
+    with Path(args.observable).open("rb") as file_obj:
+        observable: SparsePauliOp = pickle.load(file_obj)  # noqa: S301
+
+    if args.simulator:
+        backend = build_simulator_backend()
+        estimator = QDMIEstimator(backend)
+    else:
+        backend = IQMBackend(
+            base_url=args.base_url or os.getenv("IQM_BASE_URL") or "https://resonance.iqm.tech",
```
burgholzer:
```suggestion
            base_url=args.base_url,
```
No need to replicate the same fallback logic a dozen times across various call sites.

python/iqm/qdmi/estimator.py
```diff
@@ -0,0 +1,112 @@
...
+        )
+        estimator = backend.estimator()
+
+    vqe = VQE(estimator, ansatz, L_BFGS_B(maxiter=args.maxiter))
+    result = vqe.compute_minimum_eigenvalue(operator=observable)
+    optimal_params = result.optimal_parameters
+    if optimal_params is None:
+        msg = "VQE optimization did not return optimal parameters."
+        raise RuntimeError(msg)
+    print([float(value) for value in optimal_params.values()])
```
burgholzer: This should probably also return the minimum eigenvalue found. I would assume that's a quantity many people would also be interested in. In the end, that's the point of VQE.

Edit: Related to a comment in the sampler, I believe we should also just return/print a serialized version of `result` here.

python/iqm/qdmi/sampler.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def __getitem__(self, key: str) -> object:
+        """Read a field from the data bin."""
+
+
+class _SupportsSamplerPrimitive(Protocol):
+    """Protocol for sampler primitives with a run method."""
+
+    def run(self, pubs: list[tuple[object, ...]], shots: int) -> _SupportsResult:
+        """Execute sampler pubs and return a job handle."""
```
burgholzer: All of these protocols and the respective casts below feel very weird to me and neither helpful nor particularly readable. Can we get rid of these?

python/iqm/qdmi/sampler.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def build_simulator_backend() -> QDMIBackend:
+    """Create the MQT Core DDSIM backend used for simulation.
+
+    Returns:
+        The MQT Core QDMI simulator backend.
+    """
+    return QDMIProvider().get_backend("MQT Core DDSIM QDMI Device")
```
burgholzer: See other comment --> I'd prefer to inline.

python/iqm/qdmi/sampler.py
```diff
@@ -0,0 +1,192 @@
...
+    with Path(args.circuit).open("rb") as file_obj:
+        circuit = qpy.load(file_obj)[0]
+
+    if args.simulator:
+        backend = build_simulator_backend()
+        circuit_for_execution = circuit
+        sampler = cast("_SupportsSamplerPrimitive", QDMISampler(backend))
+    else:
+        backend = IQMBackend(
+            base_url=args.base_url or os.getenv("IQM_BASE_URL") or "https://resonance.iqm.tech",
```
burgholzer:
```suggestion
            base_url=args.base_url,
```
see other comment

python/iqm/qdmi/sampler.py
```diff
@@ -0,0 +1,192 @@
...
+
+    args = parser.parse_args(list(argv) if argv is not None else None)
+
+    with Path(args.circuit).open("rb") as file_obj:
+        circuit = qpy.load(file_obj)[0]
+
+    if args.simulator:
+        backend = build_simulator_backend()
+        circuit_for_execution = circuit
+        sampler = cast("_SupportsSamplerPrimitive", QDMISampler(backend))
```
burgholzer: Technically, the simulator backend also needs transpilation (it jsut happens to support a very wide range of gates), so I suppose the logic here can be simplified quite a bit by moving parts out of the branches.

python/iqm/qdmi/sampler.py
```diff
@@ -0,0 +1,192 @@
...
+            tokens_file=args.tokens_file,
+            qc_id=args.qc_id,
+            qc_alias=args.qc_alias,
+        )
+        circuit_for_execution = transpile(circuit, backend)
+        sampler = cast("_SupportsSamplerPrimitive", backend.sampler())
+
+    job = sampler.run([(circuit_for_execution,)], shots=args.shots)
+    first_pub = next(iter(job.result()))
+    counts = extract_counts(first_pub)
```
burgholzer: I am not a big fan of the logic behind this. 
This is making assumptions on the submitted circuit that might hold in many, but definitely not in all cases.
How about we simply return a serialized version of `job.result()` here (either serialized as JSON or as binary; whichever is more convenient; probably JSON).
Then the caller can reconstruct whatever they want from the result.
They also know way better what the registers they used were called etc.

test/python/test_estimator.py
```diff
@@ -0,0 +1,74 @@
...
+    params = ast.literal_eval(result.stdout.strip())
+    assert len(params) == 1
+    assert math.isfinite(float(params[0]))
+
+
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="The subprocess calls do not work properly on Windows.")
+def test_estimator_execute_module() -> None:
+    """Test running the estimator CLI by executing the iqm.qdmi.estimator module."""
+    output = check_output([sys.executable, "-m", "iqm.qdmi.estimator", "--help"])
+    assert "Estimate VQE parameters" in output.decode()
```
burgholzer: Hm. Instead of having to skip a test (on Windows) because subprocess is used, I'd rather exercise a second script runner test that invokes the `help` command.

test/python/test_main.py
```diff
@@ -61,3 +65,10 @@ def test_cli_lib_path(script_runner: ScriptRunner) -> None:
     result = script_runner.run(["iqm-qdmi", "--lib_path"])
     assert result.success
     assert str(IQM_QDMI_LIBRARY_PATH) in result.stdout
+
+
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="The subprocess calls do not work properly on Windows.")
+def test_cli_execute_module() -> None:
+    """Test running the CLI by executing the iqm.qdmi module."""
+    output = check_output([sys.executable, "-m", "iqm.qdmi", "--version"])
+    assert __version__ in output.decode()
```
burgholzer: Based on the other comment, I'd rather remove this test. The respective code path is already exercised above.

test/python/test_sampler.py
```diff
@@ -0,0 +1,60 @@
...
+    assert sum(counts.values()) == 256
+    assert set(counts) <= {"00", "11"}
+    assert counts
+
+
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="The subprocess calls do not work properly on Windows.")
+def test_sampler_execute_module() -> None:
+    """Test running the sampler CLI by executing the iqm.qdmi.sampler module."""
+    output = check_output([sys.executable, "-m", "iqm.qdmi.sampler", "--help"])
+    assert "Sample a serialized QPY circuit" in output.decode()
```
burgholzer: Same comment as in the other file applies here.

pyproject.toml
```diff
@@ -46,7 +46,8 @@ classifiers = [
 [project.optional-dependencies]
 qiskit = [
   "mqt-core~=3.6.1",
-  "qiskit[qasm3-import]>=1.1"
+  "qiskit[qasm3-import]>=1.1",
+  "qiskit-algorithms>=0.3.0"
```
burgholzer:
```suggestion
  "qiskit-algorithms>=0.4.0"
```
I think we could safely require the latest version here.

pyproject.toml
```diff
@@ -56,11 +57,14 @@ Discussions = "https://github.com/iqm-finland/QDMI-on-IQM/discussions"
...
 [project.scripts]
 iqm-qdmi = "iqm.qdmi.__main__:main"
+iqm-sampler = "iqm.qdmi.sampler:main"
+iqm-estimator = "iqm.qdmi.estimator:main"
 
 [dependency-groups]
 qiskit = [
   "mqt-core~=3.6.1",
   "qiskit[qasm3-import]>=1.1",
+  "qiskit-algorithms>=0.3.0",
```
burgholzer:
```suggestion
  "qiskit-algorithms>=0.4.0",
```

burgholzer (changes requested):
Thanks @marcelwa 🙏🏼 Very nice additions. As usual, I got a few requests for changes/polishing. Shouldn't be too bad though hopefully.

pyproject.toml
marcelwa: I wasn't sure of that because our dependency is `"qiskit[qasm3-import]>=1.1"`.

However, as you also pointed out, in the `uv.lock`, it states
```toml
[[package]]
name = "qiskit"
version = "2.4.2"
```

burgholzer: These warning filters are only really relevant for the locked versions of the dependencies. The minimums session purposely deactivates the warning-as-error flag.

marcelwa: Got it, thanks. In that case, we can savely remove them.

pyproject.toml
marcelwa: This comes from `qiskit_algorithms.optimizers.L_BFGS_B`. It initializes with `disp = False` and `iprint = -1`. But these options are deprecated as of SciPy 1.15.

python/iqm/qdmi/estimator.py
```diff
@@ -0,0 +1,142 @@
...
+        optimal_value = float(optimal_value.real)
+
+    return {
+        "eigenvalue": serialized_eigenvalue,
+        "optimal_parameters": serialized_params,
+        "optimal_point": _serialize_value(getattr(result, "optimal_point", None)),
+        "optimal_value": optimal_value,
+        "optimizer_time": getattr(result, "optimizer_time", None),
+        "cost_function_evals": getattr(result, "cost_function_evals", None),
+    }
```
burgholzer: Hm. I am not fully convinced this is the way to go.
I would have assumed (sorry for not checking earlier) that there is some builtin way for derialization and deserialization of these result datastructures built into Qiskit.
In its current form, the client side has to have some custom way of deserializing the custom serialization of the datastructure.
This does not seem particularly practical.
Can you check how this is handled in other libraries building on Qiskit and its runtime?

It probably pays off to get this right from the get go.

burgholzer (changes requested):
Thanks for another iteration @marcelwa 🙏
I am not quite sure we have found the "right" solution for the serialization/deserialization yet; see the inline comment.

python/iqm/qdmi/estimator.py
marcelwa: It seems that Qiskit supports Python's `pickle` protocol for many of its data types. In https://github.com/Qiskit/qiskit/issues/12787, it was requested to add `pickle` support for `DataBin` (inside `PrimitiveResult` and `PubResult`), which was implemented in https://github.com/Qiskit/qiskit/pull/12963. The feature is only part of the [Qiskit 2.1.0](https://github.com/Qiskit/qiskit/releases/tag/2.1.0) release. We would have to bump our dependency version from the currently supported `1.1.0` if this is the way we want to go.

Alternatively, we could ship a small support shim that monkeypatches `DataBin`, which could be as small as:
```python
import qiskit
from qiskit.primitives.containers.data_bin import DataBin

# If pickling is not supported by default, patch it
try:
    import pickle
    pickle.loads(pickle.dumps(DataBin()))
except NotImplementedError:
    # Bypass the immutable __setattr__ restriction during unpickling
    DataBin.__setattr__ = object.__setattr__
```

What do you think?

burgholzer: Hm. Pickling sounds way more reasonable than the custom solution currently in place in the PR.
If all that is needed to patch this is the lines in your comment above, I'd aim for keeping the 1.x Qiskit compatibility; including a note about this being a workaround that can be removed once the Qiskit lower bound is 2.1+.

marcelwa: Sounds good. I'll get to it tomorrow.

burgholzer (approved):
This looks much better now! 🙏🏼 let's get this in.

PR #93, opened 2026-06-19
♻️ Dockerize the SPANK tests and CI workflow

marcelwa:
This PR dockerizes the SPANK tests and the corresponding CI workflow. This allows you to run the SPANK tests locally and simplifies the CI configuration significantly. Overall, this is a much cleaner separation of concerns.


spank/test/run_docker_tests.sh
burgholzer, line 1: A lot of the preamble here should be moved to the dockerfile in my opinion.
Shouldn't the purpose of the dockerfile be to provide a "fully working" setup, that is then tested in this test file?

burgholzer (changes requested):
This generally looks great and much cleaner already. I am just not fully happy with the split between the test file and the dockerfile yet.

spank/test/run_docker_tests.sh
```diff
@@ -0,0 +1,110 @@
...
+  sudo cat /var/log/slurmctld.log || true
+  echo "=== Slurmd Log ==="
+  sudo cat /var/log/slurmd.log || true
+}
+
+# If any command fails, dump logs before exiting
+trap show_logs ERR
+
+echo "=== Starting Munge service ==="
+sudo service munge start
```
burgholzer: This should not be part of the script. Munge should already be running; if it is necessary at all for this demonstration.

spank/test/run_docker_tests.sh
```diff
@@ -0,0 +1,110 @@
...
+
+echo "=== Starting Slurmctld ==="
+sudo /usr/sbin/slurmctld || { echo "ERROR: slurmctld failed to start"; exit 1; }
+
+echo "=== Starting Slurmd ==="
+# Slurmd needs to be started as root, bound to localhost.
+sudo /usr/sbin/slurmd -N localhost || { echo "ERROR: slurmd failed to start"; exit 1; }
```
burgholzer: Are these really start commands? Why would something be started in the script? This should be running as part of the docker setup already in my opinion.

spank/test/run_docker_tests.sh
```diff
@@ -0,0 +1,110 @@
...
+  fi
+  # Force resume if the node got marked DOWN/DRAINED
+  sudo scontrol update nodename="localhost" state=resume 2>/dev/null || true
+  sleep 1
+done
+
+if ! sinfo -h -n "localhost" -o "%t" | grep -qE "idle|alloc"; then
+  echo "ERROR: Node did not become ready after 30 seconds"
+  exit 1
+fi
```
burgholzer: Also feels like this should be part of the general setup.
Once the dockercontainer is running, I would expect a "working" setup.

spank/test/run_docker_tests.sh
```diff
@@ -0,0 +1,110 @@
...
+echo "=== Activating SPANK plugin ==="
+PLUGIN_DIR=$(scontrol show config | sed -n 's/^PluginDir *= *//p')
+PLUGIN_PATH="${PLUGIN_DIR}/iqm-spank-plugin.so"
+TEST_BASE_URL="https://test.example.com"
+echo "required ${PLUGIN_PATH} iqm_base_url=${TEST_BASE_URL}" \
+  | sudo tee /etc/slurm/plugstack.conf.d/iqm-qdmi.conf
+
+# Reload Slurm configuration
+sudo scontrol reconfigure
+sleep 2
```
burgholzer: Also all feels like part of the setup of the system. Shouldn't the tests simply run on a properly configured system?

burgholzer (changes requested):
Right direction, but let's take a couple more steps there.

spank/Dockerfile
```diff
@@ -0,0 +1,109 @@
...
+# Build, install, and globally activate the SPANK plugin at image build time
+RUN cmake -S /workspace -B /workspace/build-spank-docker -DCMAKE_BUILD_TYPE=Release -DBUILD_IQM_SPANK=ON \
+    -DIQM_QDMI_SPANK_INSTALL_DIR=/usr/lib/x86_64-linux-gnu/slurm-wlm -DIQM_QDMI_SLURM_CONF_DIR=/etc/slurm && \
+    cmake --build /workspace/build-spank-docker --target iqm-spank-plugin && \
+    cmake --install /workspace/build-spank-docker --component iqm-spank-plugin
+
+RUN echo "required /usr/lib/x86_64-linux-gnu/slurm-wlm/iqm-spank-plugin.so iqm_base_url=https://test.example.com" > /etc/slurm/plugstack.conf.d/iqm-qdmi.conf
+
+# Ensure testuser owns the build outputs so they can write to it if needed
+RUN chown -R testuser:testuser /workspace/build-spank-docker
```
burgholzer: I do not fully understand why this is necessary? After the install, nothing from the build folder should be required anymore, right?

spank/Dockerfile
```diff
@@ -0,0 +1,109 @@
...
+# Copy the workspace into the container (used when building the self-contained image)
+COPY --chown=testuser:testuser . /workspace
+
+# Build, install, and globally activate the SPANK plugin at image build time
+RUN cmake -S /workspace -B /workspace/build-spank-docker -DCMAKE_BUILD_TYPE=Release -DBUILD_IQM_SPANK=ON \
+    -DIQM_QDMI_SPANK_INSTALL_DIR=/usr/lib/x86_64-linux-gnu/slurm-wlm -DIQM_QDMI_SLURM_CONF_DIR=/etc/slurm && \
+    cmake --build /workspace/build-spank-docker --target iqm-spank-plugin && \
+    cmake --install /workspace/build-spank-docker --component iqm-spank-plugin
+
+RUN echo "required /usr/lib/x86_64-linux-gnu/slurm-wlm/iqm-spank-plugin.so iqm_base_url=https://test.example.com" > /etc/slurm/plugstack.conf.d/iqm-qdmi.conf
```
burgholzer: Why not use the resonance URL here?

spank/test/run_docker_tests.sh
```diff
@@ -0,0 +1,62 @@
...
+# If any command fails, dump logs before exiting
+trap show_logs ERR
+
+echo "=== Verifying basic Slurm srun connectivity ==="
+srun --partition=debug --immediate=5 /bin/true
+
+# Running smoke tests
+TEST_BASE_URL="https://test.example.com"
+echo "=== Running SPANK smoke tests ==="
+uv run nox -f /workspace/spank/noxfile.py --envdir /home/testuser/.nox -s smoke_tests -- \
```
burgholzer: this looks fairly awkward. Especially since this will install the project.

Why is this not simply `uvx nox -s smoke_tests -- ...`?

burgholzer (changes requested):
Sorry for being so picky here. But I'd really love for these scripts and the setup to be simple. Right now it still feels a bit convoluted.

spank/Dockerfile
marcelwa: You're right. I removed it.

spank/Dockerfile
marcelwa: We talked about this above. However, to prevent this confusion in the future, I changed it to the Resonance URL.

spank/test/run_docker_tests.sh
marcelwa: Since we have two Noxfiles in the repo, we have to specify which one is meant. If that is undesired, you have to change the working directory first. I implemented that solution below.

burgholzer: Changing the working directory feels like a much cleaner solution. 👍🏻

spank/Dockerfile
burgholzer: Yeah, I remember. I'd still like to think that if we got to choose _some_ URL, we may as well choose an existing one.

.github/workflows/spank-tests.yml
```diff
@@ -17,140 +17,14 @@ jobs:
...
-      - name: Run SPANK resonance tests
-        if: env.IQM_TOKEN != ''
+      - name: Run SPANK tests in Docker
+        env:
+          IQM_TOKEN: ${{ secrets.IQM_TOKEN }}
         run: |
-          uv run nox -f spank/noxfile.py -s resonance_tests -- \
-            --partition debug \
-            --test-base-url "https://resonance.iqm.tech" \
-            --test-qc-alias "emerald:mock"
```
burgholzer, line 156: Previously, this was setting the QC alias. Now, this is not being set anymore. Was this on purpose? In the past, this has led to the "real" QPU being selected instead of the mock.

burgholzer (approved):
Looks good to me now.
Just one last clarification question.

.github/workflows/spank-tests.yml
marcelwa: We already set the `--test-qc-alias "emerald:mock"` default in `spank/noxfile.py`. Omitting it here means that the users' environment variable is respected in the Dockerized setup, with a fallback to the Nox default.

burgholzer: Ah. That makes sense. In that case, let's resolve and merge.

PR #104, opened 2026-06-29
:sparkles: Add `offloader` module to programmatically submit jobs to Slurm

marcelwa:
🤖 *AI text below* 🤖

This adds an `offloader` module so jobs can be submitted to Slurm programmatically instead of going through the CLI by hand. It wraps `srun` for both sampling and estimation, keeps the submitted circuit and job metadata around on disk so a failed run can be inspected afterwards, and reuses the same transpile settings as the existing local/simulator paths so results stay consistent regardless of where a job actually runs.


python/iqm/qdmi/__init__.py
burgholzer, line 1: Let's not re-export these in the top-level module and keep the surface of `iqm.qdmi` small.

python/iqm/qdmi/offloader.py
```diff
@@ -0,0 +1,420 @@
...
+
+    for key in ("meas", "c"):
+        with contextlib.suppress(AttributeError, KeyError, TypeError):
+            value = data[key]
+            get_counts = getattr(value, "get_counts", None)
+            if callable(get_counts):
+                return normalize_counts(get_counts())
+
+    msg = "Could not find a classical register with counts in the primitive result."
+    raise RuntimeError(msg)
```
burgholzer, line 101: Maybe it is just me, but it feels like we are repeating this functionality in many different places. Might be worth to consolidate.

python/iqm/qdmi/offloader.py
```diff
@@ -0,0 +1,420 @@
...
+    if base_url := os.getenv("IQM_BASE_URL"):
+        args.extend(["--base-url", base_url])
+
+    if tokens_file := os.getenv("IQM_TOKENS_FILE"):
+        args.extend(["--tokens-file", tokens_file])
+
+    if qc_id := os.getenv("IQM_QC_ID"):
+        args.extend(["--qc-id", qc_id])
+    if qc_alias := os.getenv("IQM_QC_ALIAS"):
+        args.extend(["--qc-alias", qc_alias])
```
burgholzer: Also kind of feels like we are repeating this in multiple places. The individual scripts that run on the offloaded node supposedly already set the environment variables (e.g. through the SPANK plugin). So can't we just eliminate this here (and potentially in other places)?

python/iqm/qdmi/offloader.py
```diff
@@ -0,0 +1,420 @@
...
+        raise ImportError(msg) from _IMPORT_ERROR
+    if local:
+        if simulator:
+            backend = QDMIProvider().get_backend("MQT Core DDSIM QDMI Device")
+            qc_for_execution = qc
+            sampler = QDMISampler(backend)
+        else:
+            backend = IQMBackend()
+            qc_for_execution = transpile(qc, backend, optimization_level=3)
+            sampler = backend.sampler()
```
burgholzer: We had this elsewhere already and decided that these paths should be unified because both actually need transpilation. Somehow this must have been lost along the way.

python/iqm/qdmi/offloader.py
```diff
@@ -0,0 +1,420 @@
...
+
+        job = sampler.run([(qc_for_execution,)], shots=shots)
+        first_pub = next(iter(job.result()))
+        return extract_counts(first_pub)
```
burgholzer, line 231: Shouldn't this just use the builtin sampler functionality that is also called remotely? Feels like repeated logic here (and below).

python/iqm/qdmi/offloader.py
```diff
@@ -0,0 +1,420 @@
...
+        Path to the jobs directory.
+    """
+    # Check if environment variable is set
+    env_dir = os.getenv("IQM_JOBS_DIR")
+    if env_dir:
+        return Path(env_dir)
+
+    # Default to user's home directory
+    default_dir = Path.home() / ".qdmi_jobs"
+    default_dir.mkdir(parents=True, exist_ok=True)
```
burgholzer: This has different semantics depending whether the Env Var is set or not. One will create the directory one will not. This should be unified so that the call site does not potentially duplicate the directory creation call as it currently does below.

python/iqm/qdmi/offloader.py
```diff
@@ -0,0 +1,420 @@
...
+        job = sampler.run([(qc_for_execution,)], shots=shots)
+        first_pub = next(iter(job.result()))
+        return extract_counts(first_pub)
+
+    # Make sure the `jobs` directory exists on the shared filesystem
+    jobs_dir = _get_jobs_dir()
+    jobs_dir.mkdir(parents=True, exist_ok=True)
+
+    job_id = _new_job_id()
+    job_dir = _make_job_dir(jobs_dir, job_id)
```
burgholzer: I can't fully remember, but shouldn't the Job ID come from SLURM?
And, in a similar line of thought, shouldn't the job directory also be provided by SLURM in some way?

python/iqm/qdmi/offloader.py
```diff
@@ -0,0 +1,420 @@
...
+
+def estimate(
+    ansatz: QuantumCircuit,
+    operator: SparsePauliOp,
+    maxiter: int = 80,
+    *,
+    local: bool = False,
+    simulator: bool = False,
+    timeout: float | None = None,
+) -> list[np.float64]:
```
burgholzer: This is not the same semantics as the estimator that is part of the regular package, is it? The other one returns a VQEResult as far as I know.

python/iqm/qdmi/offloader.py
```diff
@@ -0,0 +1,420 @@
...
+        qpy.dump(qc, f)
+
+    # Run the job using srun, which will return the counts upon completion
+    # This call is blocking and will wait for the job to finish before returning.
+    job_name = "sample_sim" if simulator else "sample_qc"
+    command = [
+        "srun",
+        f"--job-name={job_name}",
+        "--nodes=1",
+        "--partition=quantum",
```
burgholzer, line 248: This makes some assumptions on the SLURM setup that are not documented anywhere as part of this PR. In particular, the presence of a `quantum` partition (= node configured as a GRES)

burgholzer (changes requested):
Thanks @marcelwa 🙏🏼 
This one feels like it still needs work to be mergable. I hope the comments below help in that regard.

python/iqm/qdmi/offloader.py
marcelwa: Good point. Should be consolidated now.

python/iqm/qdmi/offloader.py
marcelwa: True. Also consolidated.

python/iqm/qdmi/offloader.py
marcelwa: Fixed.

python/iqm/qdmi/offloader.py
marcelwa: That was intentional. The job directory must exist and contain the serialized circuit *before* we call `srun`, so there's no Slurm job ID at that point. Happy to discuss if you see a way around that.

python/iqm/qdmi/offloader.py
marcelwa: Good point! It now returns a `VQEResult` from both the local and Slurm paths.

python/iqm/qdmi/offloader.py
marcelwa: I documented this in the docs now. Do you think that would suffice?

burgholzer: Putting it into the docs should be sufficient 👍🏼

python/iqm/qdmi/_backends.py
```diff
@@ -0,0 +1,80 @@
...
+
+if TYPE_CHECKING:
+    from mqt.core.plugins.qiskit.backend import QDMIBackend
+
+_SIMULATOR_DEVICE = "MQT Core DDSIM QDMI Device"
+
+#: Optimization level used whenever a circuit is transpiled for execution
+#: (offloader local path and the `iqm-sampler` Slurm worker), so the two
+#: paths stay in sync instead of drifting to different Qiskit defaults.
+TRANSPILE_OPTIMIZATION_LEVEL = 3
```
burgholzer:
```suggestion
TRANSPILE_OPTIMIZATION_LEVEL = 2
```
Let's stick with Qiskit's default here

burgholzer (approved):
Just one tiny comment. Then let's get this in 🚀

PR #105, opened 2026-06-29
♻️ Rewrite the HTTP client in cpr

marcelwa:
This PR rewrites the HTTP client, which previously used plain cURL, to use [cpr](https://github.com/libcpr/cpr).


marcelwa:
@burgholzer this is an initial solution to the cpr rewrite that fully passes the CI. I had to trick `clang-tidy` a bit to find the cpr library during checking. Not sure if that is the prettiest solution.

Also, there was an issue with the macOS wheels during build. I tried to essentially have all wheels simply build the cURL version packaged with cpr (instead of relying on a system-provided one), but since that one is rather new, it would have required raising the macOS deployment target to version 26 (if I remember correctly). In order to prevent that, we now have an inconsistent setup across operating systems:
```toml
[tool.scikit-build.cmake.define]
BUILD_IQM_QDMI_TESTS = "OFF"
CPR_USE_SYSTEM_CURL = "OFF"

[[tool.scikit-build.overrides]]
if.platform-system = "darwin"
cmake.define.CMAKE_IGNORE_PATH = "/opt/homebrew;/usr/local"
cmake.define.CPR_USE_SYSTEM_CURL = "ON"
```

Long story short, I'd love your input on those matters. Thanks in advance! :pray:

burgholzer:
@marcelwa now I see what you meant with the macOS problems.
Unfortunately, this is quite a common problem with brew-provided libraries on macOS. These are always built for the native machine without any kind of backwards compatibility.
I think the main problem here are the OpenSSL libraries that are being pulled in.
These would have to be built from source similar to the cpr and curl libraries for this to become a standalone executable.

burgholzer:
https://github.com/iqm-finland/QDMI-on-IQM/pull/105/commits/ab74ab9c14392fa8fb44341abac8cd5db5a59c72 addresses the concerns I had in the past on the implementation of the HTTP client leaking just because of testing concerns. The refactored design is much simpler and cleaner in my opinion.
The shared library of the QDMI device now only exposes the symbols it ought to expose.

I am pretty sure that clang-tidy will complain quite a bit on some of the naming, but this should be easily fixable.

pyproject.toml
```diff
@@ -114,6 +114,11 @@ sdist.exclude = [
 [tool.scikit-build.cmake.define]
 BUILD_IQM_QDMI_TESTS = "OFF"
 
+[[tool.scikit-build.overrides]]
+if.platform-system = "darwin"
+cmake.define.CMAKE_IGNORE_PATH = "/opt/homebrew;/usr/local"
```
burgholzer, line 119: What's the purpose of this line here?

burgholzer (approved):
I believe this would be ready to go in. I just have one question for clarification.
Overall this is a great improvement 🙌

pyproject.toml
marcelwa: This makes `find_package(CURL)` skip the Homebrew install locations when `CPR_USE_SYSTEM_CURL=ON`.

Without that line, the CI runners would always pick up the Homebrew cURL and thereby also Homebrew's OpenSSL. That's exactly what we don't want because if we built our wheels against that, they couldn't safely be repaired, I'd assume.

burgholzer: Hm. That kind of makes sense. But the build succeeding also means there is another installation on the system itself.
So could we just specify this ignore setting, but rely on the cURL that ships with CPR?

marcelwa: I don't think that would work. I tried that before and then had the issue that linking against the pulled-in OpenSSL would require raising the deployment target. If I understood you correctly, you also wanted to avoid building OpenSSL from scratch.

burgholzer: Okay, briefly tried that in https://github.com/iqm-finland/QDMI-on-IQM/pull/105/commits/26a20e43ccc042ecab6e495966e72bde33b6d45f, but it failed as you expected due to a mixture of shared libs being picked up from the system.
I reverted that commit and this should now be ready to go in.

PR #114, opened 2026-07-06
:sparkles: Add Slurm License-based Capacity Limiting to the SPANK plugin

marcelwa:
This PR adds optional Slurm license-based capacity limiting to the SPANK plugin, so admins can limit how many jobs concurrently target a given QPU. This could be useful for Resonant calls as well (they have their own queuing system anyway), but is especially important for on-premises ones, which are probably single-tenant hardware with no cloud-side queue to absorb overlapping requests.

When a job resolves `IQM_QC_ALIAS`, the plugin checks its Slurm license request (`--licenses=iqm_qc_<alias>:1`) against a configurable expected name, and either warns or, with `iqm_require_license=1`, fails the job.

While implementing this, I also found and fixed a pre-existing bug in which the plugin's validation checks logged errors but never actually blocked the task, because `ESPANK_ERROR` is a positive value and Slurm's task launcher treats a negative hook return as fatal.


burgholzer:
@marcelwa I haven't yet looked at the PR itself, but would you mind extracting the bugfix into a separate PR that we can quickly get in while we discuss about this one here?

marcelwa:
> @marcelwa I haven't yet looked at the PR itself, but would you mind extracting the bugfix into a separate PR that we can quickly get in while we discuss about this one here?

Sure, done in https://github.com/iqm-finland/QDMI-on-IQM/pull/117.

docs/spank_plugin.md
```diff
@@ -76,6 +76,25 @@ To run this job:
...
+srun --partition=quantum --iqm-qc-alias=emerald --licenses=iqm_qc_emerald:1 python bell_state.py
+```
+
+This matters most for **on-premise QCs**: unlike the cloud Resonance API, which
+queues overlapping requests on its own, an on-premise QC is typically
+single-tenant hardware with no such queue, so Slurm itself must serialize
+access. If you omit `--licenses` on a cluster where it's expected, the plugin
+logs a warning (or, if the administrator has set `iqm_require_license=1`, fails
+your job step at launch instead — after it has already been allocated, not at
+submission time).
```
burgholzer: While this might be true in general, this is not quite true for an on-premise IQM setup (at the moment), which still runs an internal queueing system just like on Resonance.

This may prompt some adjustments in the documentation wording here.
Licenses may still be useful as they allow one to kind of regulate the internal queue pressure.

docs/spank_plugin.md
```diff
@@ -156,12 +192,59 @@ After modifying the configuration, apply changes to the cluster:
...
+capacity-limiting scheduler resource — unrelated to the GPLv3/Apache-2.0
+software licensing described elsewhere in this repository.
+:::
+
+Each QC can be modeled as a flat, cluster-wide Slurm license so that Slurm
+itself enforces a concurrency limit, rather than relying on jobs to behave. This
+is especially important for **on-premise QCs**: unlike the cloud Resonance API,
+which absorbs overlapping requests in its own queue, an on-premise QC is
+typically single-tenant hardware, so uncontrolled concurrency risks real
+hardware contention.
```
burgholzer: Similar to above, I believe this requires rewording.

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ releases may include breaking changes.
 
 ### Added
 
+- ✨ Validate Slurm `--licenses` alignment with the targeted QC alias in the
+  SPANK plugin, enabling admins to enforce Slurm-native concurrency limits on
+  on-premise QCs ([#114]) ([**@marcelwa**])
```
burgholzer, line 17: Similar to my comment in the other PR, I believe it is reasonable to simply add this PR to the entry for the SPANK plugin itself.

burgholzer (changes requested):
Thanks @marcelwa 🙏🏼 
Besides the conflicts that need to be resolved, I only have a high level comments here. Then this can happily go in.

docs/spank_plugin.md
marcelwa: Reworded — on-premise IQM setups do still run their own internal queue (like Resonance), they just typically front single-tenant hardware. Reframed this as `Slurm regulating queue pressure` rather than `no queue at all`. Same fix applied to the mirrored explanation below and to the equivalent comment in `iqm_spank_plugin.cpp`.

docs/spank_plugin.md
marcelwa: Fixed with the same reword as above.

CHANGELOG.md
marcelwa: Since v1.2.0 shipped in the meantime without this PR, I couldn't fold it into the now-released SPANK plugin entry the way #117 was folded — kept it as its own bullet under [Unreleased] instead, and dropped the separate Fixed entry (it was just restating the already-released #117 fix with the new license clause tacked on).

burgholzer (approved):
LGTM. Let's get this in.

PR #115, opened 2026-07-06
🔧 Maintenance round

denialhaag:
This PR makes several maintenance updates:

- Set up `rumdl`
- Use `run-python-linter` output
- Fix `required-checks-pass` job
- Update `.pre-commit-config.yaml`
- Make use of the fact that version 4.0.0 of `cbuildwheel` installs `delvewheel` by default


spank/LICENSE.md
burgholzer, line 1: I suppose the changes here should be reverted, and this should be turned into a plain file as well to be consistent.

burgholzer (changes requested):
Thanks @denialhaag 🙏🏼 This looks great except for one small oversight that should be easy to fix.

spank/LICENSE.md
denialhaag: That was actually intentional because that license is a proper Markdown document with headings and everything. 🤔

burgholzer: Ah, nvm then. Should have checked more carefully! 👍🏼  Resolving and merging.

burgholzer (approved):

PR #117, opened 2026-07-07
🐛 Actually block job launch on validation failure in SPANK plugin

marcelwa:
## Summary

Split out of #114 at [@burgholzer's request](https://github.com/iqm-finland/QDMI-on-IQM/pull/114#issuecomment-3346103524), so this pre-existing bugfix can land independently of the Slurm license capacity-limiting feature.

`slurm_spank_task_init()` returned `ESPANK_ERROR` directly on validation failure, but `ESPANK_ERROR` is `+3000` and slurmstepd's task-init hook (`task.c`, `spank_user_task(...) < 0`) only treats a *negative* return as fatal. As a result, jobs with invalid IQM configuration (missing `IQM_BASE_URL` or conflicting `IQM_TOKEN`/`IQM_TOKENS_FILE`) were logged as errors but launched anyway instead of being blocked.

This PR translates any internal validation failure to a plain `-1` before returning from the hook, so a `required` plugin actually blocks the task launch as intended.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ releases may include breaking changes.
 
 ### Fixed
 
+- 🐛 Fix the SPANK plugin silently launching jobs despite failed validation
+  (missing `IQM_BASE_URL` or conflicting auth), since `ESPANK_ERROR` is a
+  positive value and Slurm's task launcher only aborts on a negative hook return
+  ([#117]) ([**@marcelwa**])
 - 🩹 Ensure the QDMI device can handle devices with computational resonators
   ([#107]) ([**@burgholzer**])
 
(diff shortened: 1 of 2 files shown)


marcelwa:
Blocked by #118: the required `🚦 Check` job here is failing due to a pre-existing negation bug in `.github/workflows/CI.yml`'s `allowed-skips` computation (introduced in #115), unrelated to this PR's changes. See #118 for the fix and root-cause analysis.

Edit: no longer blocked.

CHANGELOG.md
```diff
@@ -23,6 +23,10 @@ releases may include breaking changes.
 
 ### Fixed
 
+- 🐛 Fix the SPANK plugin silently launching jobs despite failed validation
+  (missing `IQM_BASE_URL` or conflicting auth), since `ESPANK_ERROR` is a
+  positive value and Slurm's task launcher only aborts on a negative hook return
+  ([#117]) ([**@marcelwa**])
```
burgholzer: I'd rather simply add the PR number to the changelog entry that introduces the spank plugin.
Since the plugin has not been released, there is little point in an entry saying "introduce spank" followed by an entry "fixed something in spank" in my opinion.
Sorry for being so picky here 😅

burgholzer (approved):
LGTM modulo one note on the changelog

CHANGELOG.md
marcelwa: Sure, done :white_check_mark:

PR #130, opened 2026-07-22
🩹 Time out cleanly and fail clearly on empty results in the offloader

marcelwa:
🤖 *AI text below* 🤖

Now that #104 has merged, this branch has been rebased onto `main` (base already updated by GitHub); the diff is just the two commits below.

These two issues were found during adversarial review of #104's fix commits, but neither is actually related to what that PR's review comments were about, so they're split out here rather than padding out an already-large review:

- `_run_srun()` parsed and passed through a `--timeout` argument that never did anything: `IQMBackend` (and the `add_dynamic_device_library()` call it wraps) has no timeout parameter, so the value could never reach anything that used it. Repurposed it to bound the `srun` subprocess call itself, so a hung Slurm job now fails with a clear `RuntimeError` instead of blocking forever. Dropped the now-pointless `--timeout` flag from the `iqm-sampler`/`iqm-estimator` worker CLIs, since it never had any effect there.
- `sample()` extracted the first pub result via a bare `next(iter(...))`, which raises an unhelpful `StopIteration` if a job unexpectedly returns no pubs. Added a small helper that raises a `RuntimeError` explaining what happened instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@ releases may include breaking changes.
 ### Added
 
 - ✨ Add `iqm.qdmi.offloader` module exposing programmatic `sample` and
-  `estimate` functions for Slurm job submissions ([#104]) ([**@marcelwa**])
+  `estimate` functions for Slurm job submissions ([#104], [#130])
+  ([**@marcelwa**])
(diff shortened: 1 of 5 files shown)


burgholzer (approved):
LGTM 👍🏼 
Please add the PR number to the respective changelog entry for the offloader before merging the PR.

PR #133, opened 2026-07-23
:sparkles: Support selecting a specific QC via SPANK from the offloader

marcelwa:
🤖 *AI text below* 🤖

Lets a caller pin a specific quantum computer for a single offloaded job. The `offloader` module added in #104 can already sample and estimate on Slurm, but had no way to target one QC among several — `sample()` and `estimate()` now accept `qc_id`/`qc_alias`, forwarded as `--iqm-qc-id`/`--iqm-qc-alias` on the `srun` command line, which the SPANK plugin resolves into the job's `IQM_QC_ID`/`IQM_QC_ALIAS` environment variables.

Both flags are only meaningful on the non-local (Slurm) path and default to unset, so existing callers see no behavior change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ releases may include breaking changes.
 - ✨ Validate Slurm `--licenses` alignment with the targeted QC alias in the
   SPANK plugin, enabling admins to enforce Slurm-native concurrency limits on
   on-premise QCs ([#114]) ([**@marcelwa**])
+- ✨ Add `qc_id`/`qc_alias` SPANK device-selection parameters to
+  `iqm.qdmi.offloader.sample`/`estimate` ([#133]) ([**@marcelwa**])
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ releases may include breaking changes.
 - ✨ Validate Slurm `--licenses` alignment with the targeted QC alias in the
   SPANK plugin, enabling admins to enforce Slurm-native concurrency limits on
   on-premise QCs ([#114]) ([**@marcelwa**])
+- ✨ Add `qc_id`/`qc_alias` SPANK device-selection parameters to
```
burgholzer: Please combine this entry with the already existing and not yet released offloader entry.

Potentially worth to update the AGENTS.md instructions in the repository so that we can more often avoid comments like these in the future.

burgholzer (approved):
LGTM except for minor nitpicks on the changelog.

burgholzer (approved):
👍🏼

PR #148, opened 2026-07-29
:bug: Guard missing VQE optimal parameters

### LINKED ISSUE #135: offloader.estimate()'s local path doesn't guard against VQEResult.optimal_parameters being None
🤖 *AI text below* 🤖

`qiskit_algorithms.VQE.compute_minimum_eigenvalue`'s result type, `VQEResult`, types `optimal_parameters` as `dict[Parameter, float] | None` — it can genuinely be `None` if the optimizer fails to converge to any parameters. `iqm.qdmi.offloader.estimate()`'s `local=True` path returns this result object directly without checking for that case, so a caller doing `result.optimal_parameters.values()` gets an `AttributeError` on `None` rather than a clear error.

burgholzer:
🤖 *AI text below* 🤖

## Summary

- validate VQE results returned by both local and Slurm estimation paths
- raise a clear RuntimeError when no optimal parameters are available
- cover both failure paths with hermetic regression tests and document the behavior

Closes #135

## Verification

- `uvx nox -s tests-3.14` (30 passed, 3 credential-gated live tests skipped)
- `uvx nox -s lint`
- `uvx nox -s docs -- -D nb_execution_mode=off` (passes with 19 existing-style warnings)

The regular documentation build reached the changed guide, then attempted an unrelated live IQM example in `docs/qiskit.md` and failed with HTTP 401 because no hardware credentials were provided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ releases may include breaking changes.
   Slurm job step before launching tasks ([#136]) ([**@burgholzer**])
 - ✨ Add `iqm.qdmi.offloader` module exposing programmatic `sample` and
   `estimate` functions (including `qc_id`/`qc_alias` SPANK device-selection
-  parameters) for Slurm job submissions ([#104], [#130], [#133])
+  parameters) for Slurm job submissions ([#104], [#130], [#133], [#148])
   ([**@marcelwa**])
 - ✨ Validate Slurm `--licenses` alignment with the targeted QC alias in the
   SPANK plugin, enabling admins to enforce Slurm-native concurrency limits on
@@ -125,6 +125,7 @@ Compatible with QDMI `v1.3.0`.
 
 <!-- PR links -->
 
+[#148]: https://github.com/iqm-finland/QDMI-on-IQM/pull/148
 [#147]: https://github.com/iqm-finland/QDMI-on-IQM/pull/147
 [#136]: https://github.com/iqm-finland/QDMI-on-IQM/pull/136
 [#134]: https://github.com/iqm-finland/QDMI-on-IQM/pull/134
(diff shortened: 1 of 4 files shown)


burgholzer:
@marcelwa I had Codex tackle the open issue, which produced the PR here.
However, I have now thought about this a little more and I do not see how this would actually be necessary. The offloader promises to return a `VQEResult`; and it does.
There is no guarantee that the optimizer converged. It is a user responsibility to check the optimal parameters are not `None`. A RuntimeError is quite confusing here actually.

I would propose to close this PR, as well as close the respective issue as "not planned".
What is your opinion?

marcelwa:
> @marcelwa I had Codex tackle the open issue, which produced the PR here. However, I have now thought about this a little more and I do not see how this would actually be necessary. The offloader promises to return a `VQEResult`; and it does. There is no guarantee that the optimizer converged. It is a user responsibility to check the optimal parameters are not `None`. A RuntimeError is quite confusing here actually.
> 
> I would propose to close this PR, as well as close the respective issue as "not planned". What is your opinion?

I try to wrap my head around this. The guard you're proposing to drop was literally implemented in the Aurora demonstrator's offloader, right? Would you argue that it was already a poor design choice back then? 

I don't feel particularly strongly about where the check lives. Caller-side is a default for many APIs. But it only works if it is documented properly that `optimal_parameters` can be `None`.

burgholzer:
> Would you argue that it was already a poor design choice back then?

Yes. (and no; Aurora previously only returned the optimal parameters and not the VQEResult; as such, failing when no optimal parameters are available was reasonable then. Now that we return the full `VQEResult`, there is no reason to fail)

> But it only works if it is documented properly that `optimal_parameters` can be `None`.

This is what the VQEResult documentation already does.

---

I am proceeding to close this and the respective issue.

PR #163, opened 2026-08-07
⚡️ Cache QDMI-on-IQM HTTP connections

burgholzer:
🤖 *AI text below* 🤖

## Summary

- add one CPR connection pool to each opaque QDMI device session
- keep request-local `cpr::Session` objects while sharing the owning device session's connection cache
- reuse the same pool across initialization, architecture refreshes, job submission and polling, result retrieval, cancellation, calibration requests, and rate-limit retries
- preserve existing authentication, headers, bodies, timeout budgets, retry behavior, logging, and response mapping
- cover pool propagation, retry reuse, five-request initialization reuse, and isolation between separately allocated device sessions

## Motivation

QDMI device initialization performs five sequential HTTP requests. Creating and destroying an independent CPR session for every request also discarded libcurl's connection cache, so those requests could each establish a separate TCP/TLS connection. A per-device `cpr::ConnectionPool` allows opportunistic keep-alive reuse without making CPR sessions persistent or introducing cross-device global state.

This follows the connection behavior measured in [openQSE/QFw#34](https://github.com/openQSE/QFw/pull/34).

## Validation

- focused pool and HTTP tests: 17 passed
- `ctest -C Release --test-dir build/test/unit --output-on-failure` (68 passed)
- `uvx nox -s lint`
- `git diff --check`
- fresh independent read-only verification

No live IQM hardware or credential-dependent integration tests were run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ releases may include breaking changes.
   Slurm license a site administrator may require via the SPANK plugin's
   `iqm_require_license` option ([#162]) ([**@marcelwa**])
 
+### Changed
+
+- ⚡️ Reuse HTTP connections within each QDMI device session to reduce TCP/TLS
+  setup during initialization and subsequent requests ([**@burgholzer**])
+
 ## [1.3.0] - 2026-07-31
 
 ### Added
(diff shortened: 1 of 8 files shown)


burgholzer:
Ignoring the QIR failure which is very likely a Resonance-internal bug according to the analysis in https://github.com/iqm-finland/QDMI-on-IQM/pull/160#issue-5073111117

PR #173, opened 2026-08-09
👷 Run the C++ test suite under AddressSanitizer and UndefinedBehaviorSanitizer

marcelwa:
🤖 *AI text below* 🤖

The C++ device does a fair amount of manual memory work — `memcpy`/`strncpy` inside the property macros, and buffer sizes computed by hand in the result getters — but nothing in CI exercises any of it under a sanitizer. Today a memory error on one of those paths first shows up as a crash inside a consumer, a long way from where it was introduced.

This adds an `IQM_QDMI_SANITIZERS` CMake option and a `cpp-sanitizers` job that uses it, wired into `CI.yml` next to the other C++ jobs and included in `required-checks-pass`. A local run is:

```console
cmake -S . -B build -DCMAKE_BUILD_TYPE=Debug -DIQM_QDMI_SANITIZERS="address;undefined"
cmake --build build
ctest --test-dir build --output-on-failure
```

Expressing this in CMake rather than as raw flags in the workflow keeps the toolchain differences in one place: MSVC ships only AddressSanitizer, spells it `/fsanitize=address`, and needs `/RTC1` stripped from its Debug flags; MemorySanitizer does not exist on macOS; ThreadSanitizer and MemorySanitizer each need their own build, since neither can share a binary with AddressSanitizer. Unsupported combinations fail at configure time with an explanation instead of at compile or link time. The flags are applied before the external dependencies are declared so those are instrumented too, which avoids the false reports that come from an allocation crossing between instrumented and uninstrumented code.

Two deliberate departures from the regular test workflows:

- `-fno-sanitize-recover=all` (via `IQM_QDMI_SANITIZER_HALT_ON_ERROR`, on by default), so an UndefinedBehaviorSanitizer finding exits non-zero rather than printing a warning into an otherwise green log.
- No `--repeat until-pass:3`. A sanitizer diagnostic is deterministic, so retrying until pass would only hide it.

## What it found

The job failed on its first run, which was the point of opening it. Three tests, all genuine, all in this repository's own code — no third-party noise, so no suppressions file is needed.

| Test | Leaked | Cause |
| :--- | :----- | :---- |
| `DeviceJobMockTest.MalformedCircuitHandling` | 18 B | setting the job program twice overwrote `program_` without freeing the first buffer |
| `QDMIIntegrationTest.JobCancellation` | 890 B in 3 allocations | job never freed |
| `QDMIIntegrationTest.OptionalJobParameters` | 1333 B in 8 allocations | job never freed |

## Rebased onto `main` (2026-08-10)

This branch was conflicting after #159, #160, #172, and #174 landed. It has been rebased onto `3573bca`, and **two commits were dropped** rather than replayed:

- `13f8f66` (release the previous program buffer) — superseded. #159 replaced the raw `void *program_` with a `std::string`, which fixes that leak structurally, so this branch no longer needs its one-line `delete[]`. This is the outcome the original note here predicted: whichever landed second should take #159's version, and #159 landed first.
- `3977a78` (changelog entry for that leak fix) — dropped with it, since #159 carries its own entry.

The other two leak fixes are still needed and still on the branch. `OptionalJobParameters` gained a free on its `GTEST_SKIP` path on `main` in the meantime; the free this branch adds is on the happy path, after the last use of the job, so the two do not collide.

Verified after the rebase: a fresh `Debug` build with `-DIQM_QDMI_SANITIZERS="address;undefined"` on clang 22 runs the hermetic unit suite **89/89 green**, `MalformedCircuitHandling` included — which is the direct confirmation that dropping `13f8f66` was safe.

## Notes for review

- **`21b13f3` goes slightly beyond what CI reported.** `FoMaC::submit_job` calls `throw_if_error` after every parameter it sets, so an early failure leaked the job regardless of what the caller did. CI never hit it because the happy path was taken, but fixing only the two call sites would have left that path leaking.
- **One hazard left deliberately unfixed.** In `OptionalJobParameters`, `ASSERT_EQ(status, QDMI_JOB_STATUS_DONE)` sits between submission and the release, so a failing run reports a leak on top of the assertion failure. The same shape exists in the tests that already freed their jobs. The release is placed directly after the last use of the job so the trailing assertion cannot skip it, but fixing the class properly means a guard in the tests too — happy to do that separately.
- **Adding the job to `required-checks-pass`** means a future leak blocks merges. That is intentional, but it is a one-line change to drop it from the required list if you would rather it start out advisory.
- **The job runs the full `ctest` suite**, including the live-backend integration tests, matching what the other C++ workflows do.
- **Plain `address,undefined` is deliberate.** A local experiment adding `integer`, `implicit-conversion`, `local-bounds`, and `nullability` produced only libstdc++-internal reports (`basic_string.h` size arithmetic, `ext/atomicity.h` refcount decrements) attributed to project call sites through inlining — no project defects, and it additionally needs `-Wno-pass-failed` to get a cpr translation unit to compile.
- **The change-detection gap noted in the first comment is still open** and still not addressed here.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -115,6 +115,14 @@ jobs:
     secrets:
       IQM_TOKEN: ${{ secrets.RESONANCE_API_KEY }}
 
+  cpp-sanitizers:
+    name: 🇨 Sanitizers
+    needs: change-detection
+    if: fromJSON(needs.change-detection.outputs.run-cpp-tests)
+    uses: ./.github/workflows/cpp-sanitizers.yml
+    secrets:
+      IQM_TOKEN: ${{ secrets.RESONANCE_API_KEY }}
+
   cpp-coverage:
     name: 🇨 Coverage
     needs: change-detection
@@ -201,6 +209,7 @@ jobs:
       - cpp-tests-macos
       - cpp-tests-windows
       - cpp-install-tests
+      - cpp-sanitizers
       - cpp-coverage
       - cpp-linter
       - python-linter
@@ -216,7 +225,7 @@ jobs:
           allowed-skips: >-
             ${{
               !fromJSON(needs.change-detection.outputs.run-cpp-tests)
-              && 'cpp-tests-ubuntu,cpp-tests-macos,cpp-tests-windows,cpp-install-tests,cpp-coverage,' || ''
+              && 'cpp-tests-ubuntu,cpp-tests-macos,cpp-tests-windows,cpp-install-tests,cpp-sanitizers,cpp-coverage,' || ''
             }}
             ${{
               !fromJSON(needs.change-detection.outputs.run-cpp-linter)
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -25,6 +25,7 @@ jobs:
         .github/workflows/cpp-coverage.yml
         .github/workflows/cpp-linter.yml
         .github/workflows/cpp-install-tests.yml
+        .github/workflows/cpp-sanitizers.yml
       additional-cd-files: |
         spank/**
         .github/workflows/spank-tests.yml
--- a/.github/workflows/cpp-sanitizers.yml
+++ b/.github/workflows/cpp-sanitizers.yml
@@ -17,18 +17,12 @@ permissions:
 
 jobs:
   cpp-sanitizers:
-    name: 🧪 AddressSanitizer + UndefinedBehaviorSanitizer
+    name: 🧪 ASan + UBSan
     runs-on: ubuntu-24.04
     env:
       CMAKE_BUILD_PARALLEL_LEVEL: 4
       CTEST_PARALLEL_LEVEL: 4
       FORCE_COLOR: 3
-      # -fno-sanitize-recover promotes UndefinedBehaviorSanitizer diagnostics
-      # from a printed warning to a non-zero exit. Without it a UBSan finding
-      # leaves the test green and goes unnoticed.
-      SANITIZER_FLAGS: >-
-        -fsanitize=address,undefined -fno-sanitize-recover=all
-        -fno-omit-frame-pointer -g -O1
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - name: Install Dependencies
@@ -48,10 +42,7 @@ jobs:
         run: |
           cmake -S . -B build \
             -DCMAKE_BUILD_TYPE=Debug \
-            -DCMAKE_C_FLAGS="$SANITIZER_FLAGS" \
-            -DCMAKE_CXX_FLAGS="$SANITIZER_FLAGS" \
-            -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address,undefined" \
-            -DCMAKE_SHARED_LINKER_FLAGS="-fsanitize=address,undefined"
+            -DIQM_QDMI_SANITIZERS="address;undefined"
       - name: Build
         run: cmake --build build
       # No --repeat here, unlike the regular test workflows: a sanitizer
--- a/.github/workflows/cpp-sanitizers.yml
+++ b/.github/workflows/cpp-sanitizers.yml
@@ -1,0 +1,67 @@
+name: 🇨 • Sanitizers
+on:
+  workflow_call:
+    inputs:
+      iqm-qc-alias:
+        description: >
+          The alias of the IQM quantum computer to use for testing. Defaults to 'emerald:mock'.
+          This can be any valid IQM quantum computer alias.
+        default: "emerald:mock"
+        type: string
+    secrets:
+      IQM_TOKEN:
+        required: true
+
+permissions:
+  contents: read
+
+jobs:
+  cpp-sanitizers:
+    name: 🧪 AddressSanitizer + UndefinedBehaviorSanitizer
+    runs-on: ubuntu-24.04
+    env:
+      CMAKE_BUILD_PARALLEL_LEVEL: 4
+      CTEST_PARALLEL_LEVEL: 4
+      FORCE_COLOR: 3
+      # -fno-sanitize-recover promotes UndefinedBehaviorSanitizer diagnostics
+      # from a printed warning to a non-zero exit. Without it a UBSan finding
+      # leaves the test green and goes unnoticed.
+      SANITIZER_FLAGS: >-
+        -fsanitize=address,undefined -fno-sanitize-recover=all
+        -fno-omit-frame-pointer -g -O1
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - name: Install Dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libcurl4-openssl-dev
+      # Set up ccache for faster C++ builds
+      - name: Setup ccache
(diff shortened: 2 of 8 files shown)


marcelwa:
🤖 *AI text below* 🤖

Noting a pre-existing gap found while getting this job to run, since it is broader than this pull request and probably wants its own fix.

The first run of the sanitizer job **skipped**, and `🚦 Check` went green. The reusable change-detection filter keys off upstream's own file names:

```text
.github/workflows/reusable-cpp-tests-ubuntu.yml
.github/workflows/reusable-cpp-tests-macos.yml
.github/workflows/reusable-cpp-tests-windows.yml
.github/workflows/reusable-cpp-coverage.yml
.github/workflows/reusable-cpp-ci.yml
.github/workflows/ci.yml
```

This repository uses none of those names — its workflows are `cpp-tests-ubuntu.yml`, `cpp-coverage.yml`, and so on, and its entry point is `CI.yml`, which does not match the lowercase `ci.yml` entry either. The consequence is general: **a pull request that changes only C++ workflow files never sets `run-cpp-tests`, so every C++ job silently skips and the merge gate passes.** Editing `cpp-tests-ubuntu.yml` today gets no C++ test coverage of that edit.

The commit here registers only `.github/workflows/cpp-sanitizers.yml` via `additional-cpp-files`, which is the minimum needed for this job to be visible. Deliberately not fixing the rest in this pull request, since adding `CI.yml` and the other `cpp-*.yml` files to the filter changes which jobs run for unrelated pull requests — that seems like your call rather than something to fold in here. Happy to do it in a follow-up if you want it.

marcelwa:
🤖 *AI text below* 🤖

First real run is in, and the job failed as intended: **3 of 77 tests**, all three genuine and all three in this repository's own code. No third-party noise, so the suppressions file speculated about in the description is not needed.

| Test | Leaked | Where |
| :--- | :----- | :---- |
| `DeviceJobMockTest.MalformedCircuitHandling` | 18 B | `iqm_device.cpp:863` |
| `QDMIIntegrationTest.JobCancellation` | 890 B in 3 allocations | `iqm_device.cpp:815` via `FoMaC::submit_job` |
| `QDMIIntegrationTest.OptionalJobParameters` | 1333 B in 8 allocations | same |

The first is the known one, already fixed by #159.

The other two are new, and are test-side rather than library-side:

```text
Direct leak of 416 byte(s) in 1 object(s) allocated from:
    #1 IQM_QDMI_device_session_create_device_job  src/iqm_device.cpp:815
    #2 FoMaC::submit_job(...)
    #3 QDMIIntegrationTest_JobCancellation_Test::TestBody()

Indirect leak of 437 byte(s) in 1 object(s) allocated from:
    #1 IQM_QDMI_device_job_set_parameter  src/iqm_device.cpp:871
```

`FoMaC::submit_job` returns a raw `IQM_QDMI_Device_Job` and hands ownership to the caller; these two tests never call `IQM_QDMI_device_job_free` on it. The 416-byte direct leak is the job object, and the indirect ones are the program buffer and parameter strings it owns.

Worth noting while someone is in there: `submit_job` calls `throw_if_error` between the allocation and the return, so an early failure leaks the job as well, regardless of what the caller does. A scope guard or a `unique_ptr` with a custom deleter would fix the test leaks and that path together.

marcelwa:
🤖 *AI text below* 🤖

Green. The sanitizer job now reports **77/77 passing**, and all checks on the pull request are clean.

Fixes are in individual commits so they can be reviewed or dropped separately:

| Commit | Fix |
| :----- | :-- |
| `13f8f66` | Release the previous program buffer when the job program is set twice |
| `21b13f3` | Hold the job in a scope guard in `FoMaC::submit_job` so a failed parameter does not leak it |
| `246c15c` | Free the job in `JobCancellation` and `OptionalJobParameters` |
| `3977a78` | Changelog entry for the leak fix |
| `20b4a4b` | clang-tidy follow-up on the guard |

The review notes that were here have been folded into the pull request description instead, so they are visible without scrolling through the thread.

One thing worth knowing for anyone working on this repository: `uvx nox -s lint` does not catch what broke in `20b4a4b`. It runs `prek`, which covers clang-format but not clang-tidy; clang-tidy only runs in the `cpp-linter` job. Contributors will keep learning about that class of finding from CI rather than locally.

.github/workflows/CI.yml
```diff
@@ -19,6 +19,12 @@ jobs:
     name: 🔍 Change
     uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-change-detection.yml@579203ef22082fa5d5483412f442e50d0c5ff4a7 # v2.2.2
     with:
+      # The reusable filter recognizes upstream's own `reusable-cpp-*.yml`
+      # names, which this repository does not use, so a change to the sanitizer
+      # workflow would otherwise never set `run-cpp-tests` and the job would
+      # silently skip.
+      additional-cpp-files: |
```
burgholzer:
```suggestion
      additional-cpp-files: |
```
Feels like bloat.

.github/workflows/cpp-sanitizers.yml
```diff
@@ -0,0 +1,58 @@
+name: 🇨 • Sanitizers
+on:
+  workflow_call:
+    inputs:
+      iqm-qc-alias:
+        description: >
+          The alias of the IQM quantum computer to use for testing. Defaults to 'emerald:mock'.
+          This can be any valid IQM quantum computer alias.
+        default: "emerald:mock"
+        type: string
```
burgholzer, line 10: Is this really needed if it is never used?

cmake/Sanitizers.cmake
```diff
@@ -0,0 +1,111 @@
...
+    string(REPLACE "/RTC1" "" CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG}")
+    set(CMAKE_C_FLAGS_DEBUG
+        "${CMAKE_C_FLAGS_DEBUG}"
+        PARENT_SCOPE)
+    set(CMAKE_CXX_FLAGS_DEBUG
+        "${CMAKE_CXX_FLAGS_DEBUG}"
+        PARENT_SCOPE)
+    message(STATUS "Sanitizers enabled: address (MSVC)")
+    return()
+  endif()
```
burgholzer, line 92: Do we potentially want to simplify this and limit sanitizers to UNIX systems?

burgholzer (changes requested):
I like this very much. Just a handful of comments for slight simplifications.

cmake/Sanitizers.cmake
marcelwa: We could. But then some developers couldn't execute them on their local machines. A trade-off worth considering. I would personally keep it as is for wider support.

.github/workflows/cpp-sanitizers.yml
marcelwa: The Test step passes it through as `IQM_QC_ALIAS`, but `CI.yml` calls this workflow without a `with:` block, so it always runs the `emerald:mock` default. That mirrors `cpp-coverage.yml` and `cpp-install-tests.yml`, which declare the same input and are called with the default. I kept it for consistency with those, and so extending the sanitizer job over backends stays a one-line change in `CI.yml`. Happy to drop it if you prefer, but then it's probably worth dropping from coverage and install-tests too.

burgholzer: Hm, no strong preference. Although I agree, that consistency is key here.
Let's keep the option for now. Feel free to resolve the conversation.

cmake/Sanitizers.cmake
burgholzer: Fine with me; in the end this is code that is not touched often anyways.

burgholzer (approved):
LGTM 👍🏼

PR #180, opened 2026-08-10
🐛 Report the device as idle again once its jobs finish

marcelwa:
🤖 *AI text below* 🤖

`IQM_QDMI_device_job_submit` sets the session's device status to `QDMI_DEVICE_STATUS_BUSY`, and nothing ever sets it back. After the first submission a session reports `BUSY` through `QDMI_DEVICE_PROPERTY_STATUS` for the rest of its life, no matter what happens to the job. A consumer that waits for `IDLE` before submitting hangs.

This is currently masked downstream rather than absent: the SPANK plugin gates task launch on the device being either `IDLE` or `BUSY`, so it never notices. Anything stricter does.

## What changed

Rather than writing the status directly at submission time, the session now counts the jobs that have not reached a terminal state. A job starts occupying the device when its submission succeeds and stops when it finishes, is canceled, fails, or is freed. The device reports `IDLE` again once that count drops back to zero.

Counting is what makes the concurrent case right. Resetting to `IDLE` on the first job that reaches a terminal state — the other obvious fix — would report an idle device while a second job is still running.

Two behavior changes fall out of this that are worth calling out explicitly, because both are visible to callers:

- A submission that fails no longer marks the device busy. Previously `BUSY` was written before the request was made, so a rejected submission left the device permanently busy without a job to blame.
- Freeing a job handle that has not finished releases the device. Such a job can no longer be observed through the API, so leaving it counted would pin the session to `BUSY` with no way back. The job itself keeps running on the backend; it is the local status that is being corrected.

A job reopened through `IQM_QDMI_device_session_retrieve_device_job_by_id` is counted too when it comes back queued or running, which keeps a retrieved job consistent with a locally submitted one.

## Testing

Five unit tests over the transitions, all against the hermetic HTTP stub: a single job returning the device to idle, two concurrent jobs keeping it busy until the second finishes, a failed submission leaving it idle, a freed in-flight job releasing it, and a cancellation releasing it. All five fail against the previous behavior — I checked by reverting the source change and rerunning them.

The full unit suite is green (94/94), including a `Debug` build under `-fsanitize=address,undefined -fno-sanitize-recover=all`.

Found while reviewing the device for #173; unrelated to that PR and independent of it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ releases may include breaking changes.
 
 ### Fixed
 
+- 🐛 Report the device as idle again once its jobs reach a terminal state,
+  instead of leaving it busy for the rest of the session after the first
+  submission ([#180]) ([**@marcelwa**])
 - 🐛 Serialize move-gate and active-reset options using the canonical IQM
   RunRequest field names ([#169]) ([**@burgholzer**])
 - 🩹 Preserve exact program bytes across QDMI job parameter updates and property
@@ -163,6 +166,7 @@ Compatible with QDMI `v1.3.0`.
 
 <!-- PR links -->
 
+[#180]: https://github.com/iqm-finland/QDMI-on-IQM/pull/180
 [#177]: https://github.com/iqm-finland/QDMI-on-IQM/pull/177
 [#172]: https://github.com/iqm-finland/QDMI-on-IQM/pull/172
 [#169]: https://github.com/iqm-finland/QDMI-on-IQM/pull/169
--- a/src/iqm_device.cpp
+++ b/src/iqm_device.cpp
@@ -98,6 +98,12 @@ struct IQM_QDMI_Device_Session_impl_d {
   /// Device status
   QDMI_Device_Status device_status_ = QDMI_DEVICE_STATUS_OFFLINE;
 
+  /// @brief Number of jobs of this session that have not reached a terminal
+  ///        state yet.
+  /// @details The device is reported as busy while this is non-zero and idle
+  ///          once it drops back to zero.
+  size_t outstanding_jobs_ = 0;
+
   /// Quantum computer ID
   std::optional<std::string> quantum_computer_id_ = std::nullopt;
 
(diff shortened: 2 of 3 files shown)


burgholzer:
This is definitely something to fix, but I am not sure this is the way to fix it.
Since one of the last PRs, the device supports reading the queue length. I think we should rather make the device status depend on that than on tracking the jobs submitted through QDMI.
Like with resonance, QDMI is potentially not the only driver of a device.

What do you think?

marcelwa:
> This is definitely something to fix, but I am not sure this is the way to fix it. Since one of the last PRs, the device supports reading the queue length. I think we should rather make the device status depend on that than on tracking the jobs submitted through QDMI. Like with resonance, QDMI is potentially not the only driver of a device.
> 
> What do you think?

Thinking about it deeper, I agree with you. We're only tracking the QDMI jobs here. Any jobs submitted through other means would invalidate our statement. 

However, an empty queue does also not mean idle. It could currently be busy with a job, but without a backlog. Also, what if the device is offline or in maintenance? Wouldn't it also have an empty queue then? That doesn't mean it's available, though. So I guess relying exclusively on queue length is not the way to go either.

burgholzer:
> However, an empty queue does also not mean idle. It could currently be busy with a job, but without a backlog. 

That is true. It's still a fairly good proxy to be honest though.
Just for reference, [here](https://github.com/munich-quantum-software/amazon-braket-qdmi-device/blob/09ebf3e1d14edd0e919c8d4483617cb09d22811d/src/Device.cpp#L308-L316) is how we handled this for Amazon Braket.

> Also, what if the device is offline or in maintenance? Wouldn't it also have an empty queue then? That doesn't mean it's available, though. 

I think we could catch that through the QC health status response (which, unfortunately, is only binary).

> So I guess relying exclusively on queue length is not the way to go either.

It's better as the status quo; yet definitely not perfect, I have to agree.

marcelwa:
🤖 *AI text below* 🤖

Agreed — closing this in favor of #190, which derives the status from the server instead of from locally tracked jobs, along the lines you suggested. Health decides maintenance, queue length decides busy, and the availability windows cover the empty-queue-but-not-usable case that we were both circling.

PR #181, opened 2026-08-10
🐛 Fix a potential histogram buffer size mismatch

marcelwa:
🤖 *AI text below* 🤖

`IQM_QDMI_device_job_get_results_hist` sizes the `QDMI_JOB_RESULT_HIST_KEYS` buffer as the length of the *first* key times the number of keys, and then writes each key at its own actual length. The two only agree when every key has the same width.

The counts are held in a `std::map`, so when widths do differ the shortest key sorts first and the reported size under-counts. A caller that does exactly the right thing — query `size_ret`, allocate that many bytes, call again — gets its buffer overrun by the library.

A `{"00": 5, "111": 3}` counts payload is enough. The API reports 6 bytes and writes 7:

```text
ERROR: AddressSanitizer: heap-buffer-overflow
WRITE of size 1 at 0x7bbc6ade4816 thread T0
    #0 IQM_QDMI_device_job_get_results_hist src/iqm_device.cpp:1484:23
0x7bbc6ade4816 is located 0 bytes after 6-byte region
```

## What changed

The size is now accumulated over all the keys instead of extrapolated from one. For keys of equal length the arithmetic is identical to what it was — `n` keys of width `w` still give `n * (w + 1)` — so responses from a well-behaved backend produce byte-for-byte the same result as before.

This is a robustness fix rather than a response to an observed backend behavior: I have not seen IQM return mixed-width keys. The point is that the library should not corrupt a caller's heap if it ever does, and mixed measurement registers are a plausible way to get there.

## Testing

One unit test against the hermetic HTTP stub, covering keys of differing width. It fails on the size assertion against the previous code and trips AddressSanitizer on the write that follows.

An equal-width companion test was dropped after review: the uniform path is already exercised by several existing unit tests as well as the integration and fomac suites, so pinning its size a second time added no coverage.

Verified with a `Debug` build under `-fsanitize=address,undefined -fno-sanitize-recover=all`: the full unit suite is 91/91 green with the fix, and reproduces the overflow above without it.

Found while reviewing the device for #173.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,7 @@ releases may include breaking changes.
 
 ### Fixed
 
-- 🐛 Size the histogram key buffer from all measurement bitstrings instead of
-  the first one, which overran the caller's buffer when the keys differed in
-  width ([#181]) ([**@marcelwa**])
+- 🐛 Fix a potential histogram buffer size mismatch ([#181]) ([**@marcelwa**])
 - 🐛 Report the number of qubits without the computational resonators, which
   inflated `QDMI_DEVICE_PROPERTY_QUBITSNUM` on Star-topology devices ([#182])
   ([**@marcelwa**])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ releases may include breaking changes.
 
 ### Fixed
 
+- 🐛 Size the histogram key buffer from all measurement bitstrings instead of
+  the first one, which overran the caller's buffer when the keys differed in
+  width ([#181]) ([**@marcelwa**])
 - 🐛 Report the number of qubits without the computational resonators, which
   inflated `QDMI_DEVICE_PROPERTY_QUBITSNUM` on Star-topology devices ([#182])
   ([**@marcelwa**])
@@ -169,6 +172,7 @@ Compatible with QDMI `v1.3.0`.
 <!-- PR links -->
 
 [#182]: https://github.com/iqm-finland/QDMI-on-IQM/pull/182
+[#181]: https://github.com/iqm-finland/QDMI-on-IQM/pull/181
 [#177]: https://github.com/iqm-finland/QDMI-on-IQM/pull/177
 [#175]: https://github.com/iqm-finland/QDMI-on-IQM/pull/175
 [#172]: https://github.com/iqm-finland/QDMI-on-IQM/pull/172
--- a/src/iqm_device.cpp
+++ b/src/iqm_device.cpp
@@ -1502,7 +1502,7 @@ int IQM_QDMI_device_job_get_results_hist(IQM_QDMI_Device_Job job,
       }
     } else {
       // The keys are not guaranteed to share a length, so the buffer has to be
-      // measured from all of them. One extra byte per key covers its 
+      // measured from all of them. One extra byte per key covers its
       // separator, the last of which becomes the null terminator.
       const size_t req_size = std::transform_reduce(
           job->counts_.begin(), job->counts_.end(), job->counts_.size(),
--- a/src/iqm_device.cpp
+++ b/src/iqm_device.cpp
@@ -1502,10 +1502,8 @@ int IQM_QDMI_device_job_get_results_hist(IQM_QDMI_Device_Job job,
       }
     } else {
       // The keys are not guaranteed to share a length, so the buffer has to be
-      // measured from all of them rather than extrapolated from the first. One
-      // extra byte per key covers its separator, the last of which becomes the
-      // null terminator. For keys of equal length this is the same size the
-      // previous computation produced.
+      // measured from all of them. One extra byte per key covers its 
+      // separator, the last of which becomes the null terminator.
       const size_t req_size = std::transform_reduce(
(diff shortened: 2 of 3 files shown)


burgholzer (changes requested):
> A {"00": 5, "111": 3} counts payload is enough.

This does not make sense to me.
The bit strings have the same length by definition. There cannot be a shot that just measures two qubits and a shot that measures three.
Feels like a very fabricated cause to me.

marcelwa:
> > A {"00": 5, "111": 3} counts payload is enough.
> 
> This does not make sense to me. The bit strings have the same length by definition. There cannot be a shot that just measures two qubits and a shot that measures three. Feels like a very fabricated cause to me.

I agree with you on "this should never happen". From a purely theoretical perspective, it is an impossible scenario. However, this addition guards us against potential upstream bugs that produce malformed responses that could potentially be used in attacks. In my opinion, adding this doesn't cost us and could potentially improve security a tad.

burgholzer:
> I agree with you on "this should never happen". From a purely theoretical perspective, it is an impossible scenario. However, this addition guards us against potential upstream bugs that produce malformed responses that could potentially be used in attacks. In my opinion, adding this doesn't cost us and could potentially improve security a tad.

I would not say that this does not have a cost.
Hm. This turns an O(1) operation into a O(|shots|) operation for something that is guaranteed by-design.

That being said, this should still be negligible and I am fine with the changes to the device itself. I'd remove the regression test though because this is not an actual regression that we saw in practice and is merely hypothetical. That should shrink the PR by quite a bit.

marcelwa:
> I would not say that this does not have a cost. Hm. This turns an O(1) operation into a O(|shots|) operation

I don't fully agree with this one. The complexity grows with the number of distinct bitstrings, which in practice should be much smaller than `|shots|`. You're of course right on the asymptotic, though. However, the function wasn't O(1) to begin with. `counts_` gets populated and read on the same path. Plus, the HTTP round-trip and the JSON parse are probably the limiting factors here, to be honest.

> for something that is guaranteed by-design.

It's not our design, though. We currently read `[0]["counts"]` without any validation. Hence, I would like to keep at least the `HistogramKeysOfDifferingLength` test (which I don't think of as a regression test anyway). But I dropped the other one, whose path was covered elsewhere.

Happy to discuss further :slightly_smiling_face:

src/iqm_device.cpp
```diff
@@ -1494,8 +1495,15 @@ int IQM_QDMI_device_job_get_results_hist(IQM_QDMI_Device_Job job,
         *data_ptr = '\0';
       }
     } else {
-      const size_t bitstring_size = job->counts_.begin()->first.length();
-      const size_t req_size = job->counts_.size() * (bitstring_size + 1);
+      // The keys are not guaranteed to share a length, so the buffer has to be
+      // measured from all of them rather than extrapolated from the first. One
+      // extra byte per key covers its separator, the last of which becomes the
+      // null terminator. For keys of equal length this is the same size the
+      // previous computation produced.
```
burgholzer:
```suggestion
      // The keys are not guaranteed to share a length, so the buffer has to be
      // measured from all of them. One extra byte per key covers its 
      // separator, the last of which becomes the null terminator.
```
Reducing the AI iteration artifacts.

test/unit/test_iqm_device.cpp
```diff
@@ -1087,6 +1087,42 @@ TEST_F(DeviceJobMockTest, FullLifecycle) {
...
+
+  size_t hist_keys_size = 0;
+  ASSERT_EQ(IQM_QDMI_device_job_get_results(job, QDMI_JOB_RESULT_HIST_KEYS, 0,
+                                            nullptr, &hist_keys_size),
+            QDMI_SUCCESS);
+  // "00" + ',' + "111" + '\0'
+  EXPECT_EQ(hist_keys_size, 7U);
+
+  // Allocate exactly what the API asked for. Before the fix this reported 6 and
+  // the write below overran the buffer by one byte.
```
burgholzer:
```suggestion
  // Allocate exactly what the API asked for.
```
Iteration artifact. In the future, no one reading this will know anymore what "the fix" is supposed to mean.

CHANGELOG.md
```diff
@@ -25,6 +25,9 @@ releases may include breaking changes.
 
 ### Fixed
 
+- 🐛 Size the histogram key buffer from all measurement bitstrings instead of
+  the first one, which overran the caller's buffer when the keys differed in
+  width ([#181]) ([**@marcelwa**])
```
```suggestion
- 🐛 Fix a potential histogram buffer size mismatch ([#181]) ([**@marcelwa**])
```
I'd make this much more hypothetical. This has not been observed in practice and is merely precautionary. I'd also update the PR title accordingly.

burgholzer (changes requested):
I am fine with keeping the size computation. I'd like to tweak the wording though, which contains a few too many iteration artifacts for my taste.

marcelwa:
> I am fine with keeping the size computation. I'd like to tweak the wording though, which contains a few too many iteration artifacts for my taste.

Gladly. I definitely agree with your comments. Everything is integrated, and the title is adjusted.

burgholzer (approved):
This looks good now! Let's get it in once the CI turns green 🙌🏼

PR #182, opened 2026-08-10
🐛 Stop counting computational resonators as qubits

marcelwa:
🤖 *AI text below* 🤖

`QDMI_DEVICE_PROPERTY_QUBITSNUM` is answered with `sites_.size()`. But the site list deliberately holds qubits *and* computational resonators — they are appended together while the static architecture is parsed, which is correct for `QDMI_DEVICE_PROPERTY_SITES` and wrong for a qubit count.

On Crystal-topology devices the two happen to coincide, which is why this has gone unnoticed. On Star-topology devices (Deneb and successors) the reported qubit count is inflated by the number of resonators, so a consumer that sizes a register from `QUBITSNUM` asks for qubits the device does not have.

## What changed

The qubit count is recorded on the session while the static architecture is parsed — it is already computed there, as `architecture["qubits"].size()` — and the property is answered from that instead. The site list is untouched.

Devices without computational resonators are unaffected, since there the two counts are equal by construction.

## Testing

Two unit tests. The first adds a three-qubit, one-resonator architecture to the mock and asserts that the qubit count reports three while `QDMI_DEVICE_PROPERTY_SITES` still reports four sites; it reports four against the previous behavior. The second pins the resonator-free case so the equal-count path cannot regress.

Full unit suite green (92/92), including a `Debug` build under `-fsanitize=address,undefined -fno-sanitize-recover=all`.

Also adds a note to `docs/usage.md`. Consumers hit this distinction exactly when allocating registers, and the docs previously described `QDMI_DEVICE_PROPERTY_SITES` as "the list of available qubit sites", which reads as though the two are the same list.

Found while reviewing the device for #173.

## The integration suite was asserting the bug

Worth flagging for review, because it is the strongest evidence the fix is correct and it is the one place this PR touches an existing test's expectations.

`QDMIIntegrationTest.QuerySiteProperties` required the site list to be exactly as long as the qubit count, and every site ID to be below that count. CI failed it against `sirius:mock`, which reports **25 sites for 24 qubits** — 24 qubits plus `COMPR1`. The assertion had been passing only because the qubit count was inflated to match the site count. The test was pinning the defect.

It now asserts the real invariants: the site list is at least as long as the qubit count, and site IDs are bounded by the size of the list they actually index. On resonator-free devices both still hold with equality.

This also means the finding is not hypothetical for Deneb alone — it is observable on a mock the suite runs against today.

> **Correction.** An earlier revision of this description also named `garnet:mock` here. That was wrong: `garnet:mock` is a Crystal-topology device with no computational resonators (20 qubits, 20 sites), and it passed `QuerySiteProperties` both before and after this change. It did fail the same CI run, but on `QueuePropertiesOnGarnetMock` — an unrelated queue-position assertion. Thanks to @burgholzer for catching it. `sirius:mock` is the only mock in the matrix that exhibits the inflated count.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ releases may include breaking changes.
 
 ### Fixed
 
+- 🐛 Report the number of qubits without the computational resonators, which
+  inflated `QDMI_DEVICE_PROPERTY_QUBITSNUM` on Star-topology devices ([#182])
+  ([**@marcelwa**])
 - 🐛 Contain exceptions during device-session initialization and leave failed
   sessions retryable ([#175]) ([**@burgholzer**])
 - 🐛 Serialize move-gate and active-reset options using the canonical IQM
@@ -165,6 +168,7 @@ Compatible with QDMI `v1.3.0`.
 
 <!-- PR links -->
 
+[#182]: https://github.com/iqm-finland/QDMI-on-IQM/pull/182
 [#177]: https://github.com/iqm-finland/QDMI-on-IQM/pull/177
 [#172]: https://github.com/iqm-finland/QDMI-on-IQM/pull/172
 [#169]: https://github.com/iqm-finland/QDMI-on-IQM/pull/169
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -292,13 +292,19 @@ The following properties about the device can be queried via the
 - {cpp:enumerator}`~QDMI_DEVICE_PROPERTY_T::QDMI_DEVICE_PROPERTY_QUBITSNUM`: The
   number of qubits available on the device.
 - {cpp:enumerator}`~QDMI_DEVICE_PROPERTY_T::QDMI_DEVICE_PROPERTY_SITES`: The
-  list of available qubit sites on the device.
+  list of available sites on the device.
 - {cpp:enumerator}`~QDMI_DEVICE_PROPERTY_T::QDMI_DEVICE_PROPERTY_OPERATIONS`:
   The list of available calibrated operations on the device.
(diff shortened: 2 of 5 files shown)


burgholzer (approved):
This generally looks good to me.
Only the following sentence in the description worries me a bit
> CI failed it against sirius:mock and garnet:mock

`garnet:mock` is not a device with comp. resonators. So any such failure should be unrelated.

marcelwa:
🤖 AI text below 🤖

You're right, and thanks for pushing on it — I conflated two different failing jobs in the same CI run.

Checking the pre-fix run ([31381541803](https://github.com/iqm-finland/QDMI-on-IQM/actions/runs/31381541803), commit c775669):

| mock | architecture | `QuerySiteProperties` |
| --- | --- | --- |
| `garnet:mock` | 20 qubits, `"computational_resonators":[]` | **Passed** |
| `sirius:mock` | 24 qubits + `COMPR1` → 25 sites | **Failed** |

So `garnet:mock` is exactly as you say — no computational resonators, and it passed this test before and after. It did fail that run, but on `QueuePropertiesOnGarnetMock` (`observed_queue_position.has_value()` was false), which has nothing to do with resonators. I saw the job marked failed and wrote it up as if it were the same assertion.

`sirius:mock` is the real evidence and it holds on its own: its static architecture comes back as 24 qubits plus `COMPR1`, so 25 sites, and `QuerySiteProperties` genuinely was passing only because `QUBITSNUM` was inflated to match. The fix and the test changes are unaffected — only that sentence overstated the reach. I've corrected the description.

Separately, that `QueuePropertiesOnGarnetMock` failure looks like a flake rather than a real defect: the job appears to have completed without ever being observed in a queued state, so the queue position was never populated. Pre-existing and unrelated to this PR, but I can open an issue if you'd like it tracked.

burgholzer (approved):
LGTM 👍🏼 And thanks for clarifying!

PR #184, opened 2026-08-12
🐛 Answer the first queue-position query from the submission response

marcelwa:
🤖 *AI text below* 🤖

The response to a job submission already tells us the job's native status and, while the job is queued, its position in the queue. Both were thrown away: the status was overwritten with `QDMI_JOB_STATUS_SUBMITTED`, and the position was only formatted into a log line.

`QDMI_DEVICE_JOB_PROPERTY_QUEUEPOSITION` then re-derived over the network something the backend had reported moments earlier.

## Why this is more than a redundant request

The extra round trip is not consequence-free. `QUEUEPOSITION` calls `IQM_QDMI_device_job_check`, which can absorb an HTTP 429 `Retry-After` of tens of seconds. A job that finishes while that request is in flight is no longer queued when it returns, so the property answers `QDMI_ERROR_BADSTATE` — and the position is gone for good, even though the backend had already handed it to us at submission time.

This is observable in CI today. `QDMIIntegrationTest.QueuePropertiesOnGarnetMock` fails intermittently for exactly this reason. From the garnet job of [run 31381541803](https://github.com/iqm-finland/QDMI-on-IQM/actions/runs/31381541803), the submission response:

```json
{"id":"019feb5b-516c-…","queue_position":1,"status":"waiting",…}
```

and, in the same job, six HTTP 429s with a 30-second `Retry-After`, one of them on a job-status GET:

```text
Request to URL 'https://resonance.iqm.tech/api/v1/jobs/019feb5b-d8bf-…'
hit HTTP 429 rate limiting; retrying after 30 second(s) (attempt 1/10)
```

A 64-shot mock job finishes in well under a second, so one rate-limited poll is enough to lose the observation. The test polls up to 20 times at 100 ms intervals, but that budget is illusory: it breaks out of the loop on `QDMI_ERROR_BADSTATE`, so a single stalled request ends it. That run failed all three `ctest --repeat until-pass:3` attempts — under a rate-limited runner it is not a coin flip, it fails reliably.

Measured across the 11 completed garnet jobs since #172 merged, one failed. Small sample, and I looked at only that one failure in depth, so treat the rate as indicative rather than precise.

## What changed

The queued observation is recorded at submission time, and the first `QUEUEPOSITION` query answers from it instead of asking again.

- The value is handed out **once**. Every later query refreshes from the server exactly as before.
- Any status check in between — `job_check`, `job_wait` — supersedes it, so a fresher observation always wins.
- Submission responses that do not report the job as queued are ignored entirely; the job keeps the submitted status it had, and those paths refresh as they always did.
- The queue position is read through the existing native-status mapping rather than by repeating the `"queued"`/`"waiting"` literals, so there is one place where those names live.

## The trade-off, stated plainly

This narrows the documented contract. `docs/usage.md` said *every* property query refreshes from the server; now the first one after a queued submission does not. The cost: a caller that submits, does nothing for a long time, and then queries once gets the submission-time position rather than a current one. Its second query is correct again.

I think that is the right trade, because the submission response is the server's own statement about the job at the moment it entered the queue — for the query that immediately follows a submission it is not merely as good as a re-check, it is the observation the re-check is trying to reproduce. But it is a real behavior change and the alternative is defensible, so it is worth an explicit look. If you would rather not narrow the contract, the other option is to leave the device alone and make the integration test tolerate a missing position — cheaper, but it leaves the redundant round trip in place and the test asserting less.

## Testing

Three unit tests against the hermetic HTTP stub, all asserting on `http_stub.get_urls()` so the round trips are counted rather than assumed:

- `QueuePositionFromSubmissionSkipsTheStatusRoundTrip` — the first query returns the submission's position with **no** additional GET; the second query does issue one and returns the refreshed value. This one fails against the previous behavior.
- `StatusCheckSupersedesTheSubmissionQueuePosition` — an explicit `job_check` in between wins.
- `SubmissionQueuePositionIgnoredWhenNotQueued` — a position alongside a non-queued status is ignored and the regular refresh happens.

The last two pass against the previous behavior as well, by design: they are guards pinning that the paths this PR does not intend to change stay unchanged.

The four pre-existing queue tests from #172 pass untouched. `QueryQueuePositionRefreshesJobStatus` still exercises the refresh path genuinely, because its submission response carries neither `status` nor `queue_position`.

Full unit suite green (93/93) under a `Debug` build with `-fsanitize=address,undefined -fno-sanitize-recover=all`.

Docs updated in `docs/usage.md` and `docs/contributing.md` to describe the narrowed contract.

Found while investigating the intermittent `QueuePropertiesOnGarnetMock` failure noticed in #182.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ releases may include breaking changes.
 
 ### Fixed
 
+- 🐛 Answer the first `QDMI_DEVICE_JOB_PROPERTY_QUEUEPOSITION` query from the
+  submission response, which already reports the position, instead of losing it
+  to a status round trip that outlives the job's stay in the queue ([#184])
+  ([**@marcelwa**])
 - 🐛 Report the number of qubits without the computational resonators, which
   inflated `QDMI_DEVICE_PROPERTY_QUBITSNUM` on Star-topology devices ([#182])
   ([**@marcelwa**])
@@ -168,6 +172,7 @@ Compatible with QDMI `v1.3.0`.
 
 <!-- PR links -->
 
+[#184]: https://github.com/iqm-finland/QDMI-on-IQM/pull/184
 [#182]: https://github.com/iqm-finland/QDMI-on-IQM/pull/182
 [#177]: https://github.com/iqm-finland/QDMI-on-IQM/pull/177
 [#175]: https://github.com/iqm-finland/QDMI-on-IQM/pull/175
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -520,9 +520,10 @@ These steps correspond to the initialization sequence described in the
 **When Job Queue Position Is Queried:**
 
 - `GET_JOB_STATUS` or `GET_CALIBRATION_JOB_STATUS`: Called for every
-  `QDMI_DEVICE_JOB_PROPERTY_QUEUEPOSITION` query. The refreshed status and queue
-  position are cached together, and a value is returned only while the job is
-  queued.
+  `QDMI_DEVICE_JOB_PROPERTY_QUEUEPOSITION` query except the first one after a
+  submission that already reported a queue position. The refreshed status and
+  queue position are cached together, and a value is returned only while the job
+  is queued.
 
 **After Calibration Job Completion:**
 
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -375,9 +375,12 @@ including the
 
 After submission,
 {cpp:enumerator}`~QDMI_DEVICE_JOB_PROPERTY_T::QDMI_DEVICE_JOB_PROPERTY_QUEUEPOSITION`
-reports the number of jobs ahead of the job while it is queued. Every property
-query refreshes the job status and queue position from the IQM server. The query
-returns `QDMI_ERROR_BADSTATE` when the refreshed job is not queued and
+reports the number of jobs ahead of the job while it is queued. Property queries
+refresh the job status and queue position from the IQM server, with one
+exception: when the submission response already reported the job as queued and
+gave its position, the first query answers from that instead of asking again.
+That value is returned once; every later query goes back to the server. The
+query returns `QDMI_ERROR_BADSTATE` when the refreshed job is not queued and
 `QDMI_ERROR_NOTSUPPORTED` when the server does not provide a trustworthy queue
 position.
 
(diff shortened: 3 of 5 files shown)


burgholzer (changes requested):
I do not agree with this change.
While it makes sense at first glance, the queue position query _must_ query the API every time.
There is no strict relation between the job submission and a query for the job's position in the queue.
These are disjoint events in time. There is no guarantee that the position at query time is still the position at submission time.

So while I agree that it feels slightly wrong to discard the information of the submission call, I do not really see a way around that if one wants to avoid reporting stale data.

The flake is still worth fixing though; but likely through loosening the test checks and no library changes.

marcelwa:
🤖 *AI text below* 🤖

You are right, and the spec settles it rather than it being a judgment call. `include/qdmi/constants.h` in QDMI says that querying `QDMI_DEVICE_JOB_PROPERTY_QUEUEPOSITION` *must* refresh the job's status and queue position, and *must* return `QDMI_ERROR_BADSTATE` when the refreshed status is not `QDMI_JOB_STATUS_QUEUED`. That means `BADSTATE` is the specified answer in exactly the situation this PR was suppressing: the job really had left the queue, and the device was already reporting correctly. Amending `docs/usage.md` to describe a narrower contract was the wrong move too — that contract is not this repo's to narrow.

The flake is real and gets fixed the way you suggested, in the test: #196.

That PR keeps two pieces of this one that are spec-safe and stand on their own — adopting the submission response's native status when it says the job is queued, guarded so a terminal status is never adopted, and reading the queue position for the log line in one place instead of two. It also adds a unit test that pins the property to the server even when the submission response carries a position, so this idea cannot come back by accident. That test fails against the implementation in this PR.

Closing rather than pushing a replacement onto this branch, even though `docs/contributing.md` says not to close and reopen after a changes-requested review, because nothing of this diff survives — the successor shares no code with it and is a different change in different files. Flagging the deviation rather than hoping it passes unnoticed.

One observation, with no proposal attached: the mandated refresh costs a full `Retry-After` wait when the status request is rate limited, and that is what made the race reachable at all — six 429s with a 30-second `Retry-After` in the garnet job of [run 31381541803](https://github.com/iqm-finland/QDMI-on-IQM/actions/runs/31381541803). It is a cost of the contract, not an argument against it, and if it is ever worth revisiting it belongs in QDMI rather than here.

PR #187, opened 2026-08-12
🐛 Stop marking a still-running job as failed when the server cannot be reached

marcelwa:
🤖 *AI text below* 🤖

A single failed status request could permanently kill a healthy job. `IQM_QDMI_device_job_check` treated "we could not ask the server" as "the job failed": any non-success response other than `QDMI_ERROR_PERMISSIONDENIED` set the job to `QDMI_JOB_STATUS_FAILED` and returned `QDMI_ERROR_FATAL`. One 502, 503, or dropped connection during a multi-hour `IQM_QDMI_device_job_wait` was therefore enough to bury a job that was still running fine remotely, and the damage was permanent: the early return at the top of `job_check` short-circuits on a terminal status, so every subsequent check answered from the poisoned cache and the caller could never find out otherwise. Meanwhile the job kept occupying the site's quota. A downstream consumer hit this.

The fix is to stop letting a failed status *query* say anything about the job's status. The job status is only mutated when the server actually reports one; otherwise the mapped error code is returned and the job stays exactly as re-checkable as it was. #160 already carved `QDMI_ERROR_PERMISSIONDENIED` out for precisely this reason — a rejected read tells you nothing about the job — and this change just promotes that carve-out to the general rule, which is both the smaller diff and the more defensible semantic. `job_wait` already returns on a non-success `job_check`, so the caller still sees the error and can simply wait again afterwards; no loop change was needed.

Judgment call worth pushing back on: I included `IQM_QDMI_device_job_cancel`, which had the identical shape and, I think, the identical bug. A cancellation request the server never answered does not mean the job failed — it most likely means the job is still running, unaffected, which is the worst possible moment to tell the caller it is dead and then refuse to look again. Applying the same rule there keeps one invariant instead of two ("a failed request never mutates job status") rather than leaving a second copy of the bug behind. The counterargument is that it widens a bug fix beyond the reported symptom, so if you would rather see cancellation handled separately, say so and I will pull it back out.

The job-submission paths are deliberately untouched. There the job genuinely did fail to be created, so recording a failure is correct.

Explicitly out of scope: no retries and no backoff around the status query. Making transient failures survivable at all is the prerequisite; deciding how hard to retry is a separate piece of work with its own trade-offs.

Four regression tests drive the scripted HTTP stub through the failure modes: a dropped connection and an HTTP 502 during a status check, an HTTP 503 during a cancellation, and a dropped connection during a wait. Each asserts both halves of the property — the job keeps its previous status, and the following check actually reaches the server rather than being answered from a cached terminal status. All four fail on `main` (the job reads back as `QDMI_JOB_STATUS_FAILED` and the follow-up request is never issued) and pass here.

### DIFF
--- a/src/iqm_device.cpp
+++ b/src/iqm_device.cpp
@@ -1322,11 +1322,10 @@ int IQM_QDMI_device_job_cancel(IQM_QDMI_Device_Job job) {
       job->session_->request_timeout_);
   const auto status = iqm::http::Handle_response(job_abortion_response);
   if (status != QDMI_SUCCESS) {
-    if (status == QDMI_ERROR_PERMISSIONDENIED) {
-      return status;
-    }
-    job->status_ = QDMI_JOB_STATUS_FAILED;
-    return QDMI_ERROR_FATAL;
+    // A cancellation request that never reached the server says nothing about
+    // the job itself, which keeps running remotely. Leave the status untouched
+    // so that the job stays observable and the request can be retried.
+    return status;
   }
   job->status_ = QDMI_JOB_STATUS_CANCELED;
   LOG_DEBUG("Job cancellation response:\n" + job_abortion_response.text);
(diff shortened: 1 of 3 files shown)


src/iqm_device.cpp
burgholzer, line 1: Should this potentially log a debug message so that failures are not silently ignored?

burgholzer (approved):
LGTM besides one small clarification/suggestion.

src/iqm_device.cpp
marcelwa: Good point. Logging added.

burgholzer (approved):
LGTM and ready to merge once the conflicts are resolved and CI is green 🙌🏼

PR #188, opened 2026-08-12
🐛 Survive an IQM server schema change instead of crashing

marcelwa:
🤖 *AI text below* 🤖

`src/iqm_device.cpp` reads fields out of parsed IQM responses without checking they exist. The parsed responses are `const`, so an expression like `qc["id"].get<std::string>()` resolves to nlohmann's *const* `operator[]`, which is documented as undefined behavior for a missing key. Under a Debug build that trips nlohmann's internal assertion and aborts, which is loud and survivable. Under the shipped Release configuration `NDEBUG` compiles the assertion out and the very same expression performs an out-of-bounds read instead — silent memory unsafety in the deployed library, triggered by nothing more than IQM renaming a field. No `try`/`catch` helps here, because nothing throws.

Out-of-range *array* subscripts are the same bug without even the Debug-build warning. `connectivity[i][0]` on an edge that carries fewer than two sites, and `job_results_json_response[0]` on an empty measurement-counts array, land in `std::vector::operator[]`, where nlohmann has no assertion of its own. A fuzzing exploration over these response shapes reports `AddressSanitizer: heap-buffer-overflow, READ of size 1` for `"connectivity": [["QB2"]]` and `reference binding to null pointer` for `"connectivity": [[]]`, and reaches the second one through the public `IQM_QDMI_device_job_get_results` with a two-byte response body.

The second failure mode is the roughly dozen bare `nlohmann::json::parse(...)` calls on response bodies, and the `get<T>()` conversions that follow them. A truncated body throws `nlohmann::json::parse_error`; a field of the wrong type throws `nlohmann::json::type_error`. Both propagate straight out through the QDMI C ABI. Propagating a C++ exception through a C function pointer is undefined behavior, and in practice it calls `std::terminate` and takes the caller's process down instead of returning an error code. `IQM_QDMI_device_job_submit`, `_job_check`, `_job_cancel`, and `_job_get_results` have no handler at all today, and each has a demonstrated escape: a numeric `id` in a submission response, a non-string `status`, a non-JSON status body, and non-string `measurement_keys` or a non-numeric count in a results body.

This is not a hypothetical failure mode for this project: #169 landed because IQM renamed RunRequest fields. Schema drift has already happened here once, and the shipped configuration currently answers it with memory unsafety.

The first half of this PR replaces every unchecked read on a server response with a checked one — array subscripts as well as object keys. Required fields are read with `at()`, genuinely optional ones stay behind `contains()`/`value()`, and response bodies are parsed with the non-throwing `parse(..., nullptr, false)` overload followed by an `is_discarded()` check, the shape `Get_queue_length` already established in this file. One deliberate loosening comes with it: a calibration observation that omits `invalid` is now treated as valid rather than read out of bounds, since only observations explicitly marked invalid should be skipped. Everything else that is missing is an error.

The second half extends the exception guarding that #175 introduced on `IQM_QDMI_device_session_init` to every remaining exported entry point that performs network I/O: `IQM_QDMI_device_job_submit`, `_cancel`, `_check`, `_get_results`, and `IQM_QDMI_device_session_query_device_property`. That is where an exception can actually originate, since those are the functions that fetch a bearer token, talk to the server, and parse what comes back. The remaining entry points are deliberately left alone: `_initialize`, `_finalize`, `_session_alloc`, `_session_free`, `_session_set_parameter`, `_session_create_device_job`, `_job_free`, `_job_set_parameter`, `_session_query_site_property`, and `_session_query_operation_property` perform no I/O and no JSON handling, so their only possible exception is `std::bad_alloc` from a small allocation. `IQM_QDMI_device_job_query_property` and `IQM_QDMI_device_job_wait` reach the network only by delegating to the now-guarded `IQM_QDMI_device_job_check`, so they need nothing of their own. Guarding the allocation-only entry points too would be defensible, but it is a separate concern from schema drift and would have made this diff harder to review; happy to add it here if you would rather have the boundary be uniform.

On style: I considered a small `Guard(lambda)` helper to avoid repeating the same four-line catch chain, and decided against it. Applying it would mean either reindenting each entry point's body into a lambda, which makes the diff enormous and buries the actual change, or splitting each entry point into a thin wrapper plus a renamed implementation function, which doubles the function count. It would also leave two different idioms in the same file, since #175 and #160 already use the explicit function-try-block. As written, each guarded function costs one keyword on its signature line and six lines at the end, and the body is untouched. Push back if you would prefer the helper.

One behavioral note worth flagging: the new malformed-response branches return an error code without marking the job `FAILED`. A response we could not parse says nothing about the state of the job on the server, and the caller can retry. Existing status transitions are untouched.

The new tests fail against `main`. Most abort outright under the Debug build:

```
unit_tests: json.hpp:2174: const_reference basic_json<>::operator[](const key_type&) const:
Assertion `it != m_data.m_value.object->end()' failed.
```

The out-of-range subscript cases do not reach that assertion at all — they trip libstdc++'s container hardening instead (`Assertion '__n < this->size()' failed`), which is equally absent from a Release build. The rest escape as C++ exceptions through the QDMI entry points (`C++ exception with description "[json.exception.type_error.302] type must be string, but is number" thrown in the test body`); GoogleTest's catch-all reports those as failures, but a real C consumer gets `std::terminate` instead. `uvx nox -s lint` is clean, `clang-tidy` reports no findings on either changed file, all 104 unit tests pass, and every changed line in `src/iqm_device.cpp` is covered by the unit suite.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,8 +28,11 @@ releases may include breaking changes.
 - 🐛 Report the number of qubits without the computational resonators, which
   inflated `QDMI_DEVICE_PROPERTY_QUBITSNUM` on Star-topology devices ([#182])
   ([**@marcelwa**])
-- 🐛 Contain exceptions during device-session initialization and leave failed
-  sessions retryable ([#175]) ([**@burgholzer**])
+- 🐛 Contain exceptions during device-session initialization and at every device
+  entry point that talks to the IQM server, check every field read from a server
+  response so that a schema change or an unreadable body is reported as an error
+  instead of corrupting memory, and leave failed sessions retryable ([#175],
+  [#188]) ([**@burgholzer**], [**@marcelwa**])
 - 🐛 Serialize move-gate and active-reset options using the canonical IQM
   RunRequest field names ([#169]) ([**@burgholzer**])
 - 🩹 Preserve exact program bytes across QDMI job parameter updates and property
@@ -168,6 +171,7 @@ Compatible with QDMI `v1.3.0`.
 
 <!-- PR links -->
 
+[#188]: https://github.com/iqm-finland/QDMI-on-IQM/pull/188
 [#182]: https://github.com/iqm-finland/QDMI-on-IQM/pull/182
 [#177]: https://github.com/iqm-finland/QDMI-on-IQM/pull/177
 [#175]: https://github.com/iqm-finland/QDMI-on-IQM/pull/175
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,11 +28,7 @@ releases may include breaking changes.
 - 🐛 Report the number of qubits without the computational resonators, which
   inflated `QDMI_DEVICE_PROPERTY_QUBITSNUM` on Star-topology devices ([#182])
   ([**@marcelwa**])
-- 🐛 Contain exceptions during device-session initialization and at every device
-  entry point that talks to the IQM server, check every field read from a server
-  response so that a schema change or an unreadable body is reported as an error
-  instead of corrupting memory, and leave failed sessions retryable ([#175],
-  [#188]) ([**@burgholzer**], [**@marcelwa**])
+- 🐛 Contain exceptions so they do not escape the C interface ([#175], [#188]) ([**@burgholzer**], [**@marcelwa**])
 - 🐛 Serialize move-gate and active-reset options using the canonical IQM
   RunRequest field names ([#169]) ([**@burgholzer**])
 - 🩹 Preserve exact program bytes across QDMI job parameter updates and property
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -25,8 +25,11 @@ releases may include breaking changes.
 
 ### Fixed
 
-- 🐛 Contain exceptions during device-session initialization and leave failed
-  sessions retryable ([#175]) ([**@burgholzer**])
+- 🐛 Contain exceptions during device-session initialization and at every device
+  entry point that talks to the IQM server, check every field read from a server
+  response so that a schema change or an unreadable body is reported as an error
+  instead of corrupting memory, and leave failed sessions retryable ([#175],
+  [#188]) ([**@burgholzer**], [**@marcelwa**])
```
```suggestion
- 🐛 Contain exceptions so they do not escape the C interface ([#175], [#188]) ([**@burgholzer**], [**@marcelwa**])
```
Just to make this more concise.

burgholzer (approved):
This looks good to me and rather complete. Except for the conflicts with main and a wish for a more concise changelog entry, I have nothing to complain 😌

PR #190, opened 2026-08-14
🐛 Report the device status the shared quantum computer is actually in

marcelwa:
🤖 *AI text below* 🤖

This replaces #180, which fixed the same bug the wrong way. The bug: `IQM_QDMI_device_job_submit` set the session's device status to `QDMI_DEVICE_STATUS_BUSY` and nothing ever set it back, so after the first submission a session reported `BUSY` through `QDMI_DEVICE_PROPERTY_STATUS` for the rest of its life. A consumer that waits for `IDLE` before submitting hangs. The SPANK plugin gates task launch on either `IDLE` or `BUSY`, so it never noticed; anything stricter does.

#180 fixed that by counting the jobs this session had submitted. As @burgholzer pointed out there, that is the wrong source of truth: the quantum computer is shared, so a job submitted through the IQM REST API or through another QDMI process is invisible to such a counter, and the session would report `IDLE` while the device is running someone else's job. Wrong in the same category as the bug it fixes, just less visibly.

## What changed

The status is no longer tracked locally at all. `session->device_status_` is gone, and `QDMI_DEVICE_PROPERTY_STATUS` is answered from what the IQM Server reports at the time of the query — the same shape as the `QUEUELENGTH` property added in #172:

| What the server says | Reported status |
| --- | --- |
| `healthy: false` | `MAINTENANCE` |
| `available: []` — the queue has no availability | `MAINTENANCE` |
| `queue_length >= 1` | `BUSY` |
| otherwise | `IDLE` |

Two inputs, two endpoints: `GET /api/v1/quantum-computers/{id}/health` (new here) and the `queue-availability` endpoint that `QUEUELENGTH` already uses. The health check short-circuits, so an unhealthy quantum computer costs one request rather than two.

The threshold is 1 rather than something larger because the IQM on-demand queue is FIFO and executes one job at a time, so a single queued job already occupies the device.

Degradation is deliberate in both directions, since either endpoint is optional in practice:

- **A backend that does not expose an endpoint** (`404`) is treated as a missing signal, not as an unusable device, and keeps reporting `IDLE`. Pinning such a backend to a non-idle status would just reintroduce the bug this PR fixes for anyone on an older IQM Server.
- **A backend that fails or cannot be reached**, or that answers with a payload we cannot read, reports `BUSY`. Claiming an idle device there would invite a submission that is bound to fail. This is per query and not sticky, unlike the original bug.

## The one thing worth a second opinion

Treating an empty `available` array as `MAINTENANCE` is the part I could not verify. It follows the documented meaning of the field — the windows during which queued jobs are executed, so no windows means queued jobs will not run — and it is the analogue of the execution-window check in the Braket device you linked. But I have no live server to confirm that a quantum computer without pay-as-you-go bookings does not simply return an empty array as a matter of course. If it does, every such device would report `MAINTENANCE`, and since the SPANK gate accepts only `IDLE` or `BUSY`, it would refuse every task launch. Please sanity-check that against a real backend before merging; `BUSY` is the safer mapping if there is any doubt, and dropping the window check entirely still leaves the health and queue signals intact.

Note also that `MAINTENANCE` is a behavior change for the SPANK plugin by design: an unhealthy or unavailable quantum computer now fails backend validation instead of launching a task that would sit there.

## Testing

Eight unit tests over the mapping against the hermetic HTTP stub, including that a job submitted through this session no longer pins the device to `BUSY`. Seven of the eight fail against the previous behavior — I checked by reverting the source change and rerunning them; the eighth is the regression guard for a backend without either endpoint, which was already idle before.

The full unit suite is green (119/119), including a `Debug` build under `-fsanitize=address,undefined -fno-sanitize-recover=all`. clang-tidy reports nothing new on the touched files. The integration suite needs a live `IQM_TOKEN` and was not run locally.


burgholzer (approved):
LGTM now 👍🏼

PR #189, opened 2026-08-14
✨ Take over IQM JSON conversion from MQT Core

marcelwa:
🤖 *AI text below* 🤖

## Description

The downstream half of munich-quantum-toolkit/core#2114, which implements munich-quantum-toolkit/core#2094.

MQT Core carried `qiskit_to_iqm_json` and an IQM `MoveGate` inside its generic Qiskit adapter. Nothing in MQT Core used either — no device it ships advertises `IQM_JSON` — and this package is the only consumer, so MQT Core dropped them. This PR takes ownership.

- `python/iqm/qdmi/serializers.py` — `qiskit_to_iqm_json`, moved from MQT Core
- `python/iqm/qdmi/gates.py` — `MoveGate`
- `pyproject.toml` — advertises the serializer through MQT Core's entry point group:

  ```toml
  [project.entry-points."mqt.core.qiskit.program_serializers"]
  IQM_JSON = "iqm.qdmi.serializers:qiskit_to_iqm_json"
  ```

- `IQMBackend` supplies the MOVE gate to the Qiskit `Target` through the new `_EXTRA_GATES` hook

The entry point matters rather than an import-time registration: MQT Core's `QDMIProvider` builds a plain `QDMIBackend` for every registered device, so someone can reach a registered IQM device without ever importing `iqm.qdmi`. Our device advertises only `QIR_BASE_STRING` and `IQM_JSON` (`src/iqm_device.cpp:2129`), so without a discoverable serializer that path has no format left to serialize to.

`test/python/test_serializers.py` ports the conversion cases MQT Core dropped, against a small stub backend — the serializer only reads `backend.device.sites()` and `site.name()`, so the 300-line mock from MQT Core is not needed. One test asserts MQT Core actually resolves our serializer through the entry point.

### Requires MQT Core 3.9.1

`mqt-core` moves from `~=3.9.0` to `~=3.9.1`. 3.9.1 is the floor because it is the release that backports #2114 to the 3.x series: the `mqt.core.qiskit.program_serializers` entry point group, the `QDMIBackend._EXTRA_GATES` hook, the public `QDMIBackend.device` property, and the removal of `qiskit_to_iqm_json` and `MoveGate` from MQT Core itself.

[...]

PR #195, opened 2026-08-20
⬆️ Consume the MQT Core 3.9.0 and QDMI 1.3.3 releases

marcelwa:
🤖 *AI text below* 🤖

## Description

Takes the MQT Core 3.9.0 and QDMI 1.3.3 releases (both 2026-08-19), adopts what they newly expose, and tests the seam.

### MQT Core 3.8.0 → 3.9.0

Two of its changes break this repo.

- **`mqt.core.fomac` → `mqt.core.qdmi.driver`** (#2074). The old module still works but emits a `DeprecationWarning`, and `[tool.pytest]` sets `filterwarnings = ["error"]`. On 3.9.0 that import fails collection across the whole Python suite.
- **`QDMISampler` and `QDMIEstimator` lost their `options` mapping** in favour of explicit shot and precision defaults (#2084). `IQMBackend.sampler` and `IQMBackend.estimator` are removed rather than rewritten: the inherited methods have the same signatures, and the inherited `estimator` also accepts `default_shots`. The two callers that passed `options={"default_shots": …}` now pass `default_shots=`.

Checked rather than assumed, and not applicable here: there is no `[tool.qdmi]` table to migrate to `qdmi.json`, and no `libmqt-core-*.so.3.8` entry in any wheel-repair config, so the SOVERSION note does not bite. `MQT::CoreFoMaC` and the C++ `fomac::` namespace are unchanged in 3.9.0, so `docs/usage.md` keeps that target name.

### QDMI → the released v1.3.3 tag

`QDMI_REV` pointed at a `develop` commit that predates the release. The tag is 13 commits ahead of it and all 13 are maintenance — renovate, lockfiles, docs, pre-commit sync — so the interface is identical and this is a provenance fix. `#485`, `#486`, and `#475` were already in the pinned commit.

1.3.3 also adds `configure_qdmi_device_target`, which replaces the block in `src/CMakeLists.txt` that set `QDMI_DEVICE_ID` and `QDMI_DEVICE_PREFIX` and appended them to `EXPORT_PROPERTIES` by hand. The function body is those same two calls plus argument validation.

I did not adopt the template's overridable `<PREFIX>_QDMI_DEVICE_ID` cache variable. `iqm.default` is also hard-coded in `python/iqm/qdmi/__init__.py`, and a settable CMake ID would let the two drift.

### New coverage

The device already reports its queue length and a queued job's position (#172), retrieves an existing job by ID (#160), and accepts calibration jobs. MQT Core 3.9.0 is the first release whose Python API reaches all of them, so `test/python/test_device_capabilities.py` covers that seam. The device internals are already covered under `test/unit/`.

**No test submits a calibration run.** That is a real operation on a shared quantum computer, and CD passes `IQM_TOKEN` into the wheel tests, so a live call would fire on every wheel job. The tests assert only that the device advertises `ProgramFormat.CALIBRATION`.

`QDMIBackend.device` is private in 3.9.0, so the tests reach the device through `mqt.core.qdmi.driver.open_device` — the same call `IQMBackend.__init__` makes.

`docs/python_package.md` gains a section on all four, with a warning on the calibration entry point.

### Dead pyright pragmas removed

Its own commit, so it is easy to drop. `test/python/test_qiskit_backend.py` carried a `# pyright: report…=false` header, and I copied it into the new test file before noticing that this project type-checks with `ty` and has no pyright configuration anywhere. Both headers are gone and `ty` passes without them.

## Two findings, no change requested

Neither is a defect in this repo. Both are recorded here rather than papered over.

1. **PennyLane cannot reach an IQM device.** MQT Core 3.9.0 added a PennyLane QDMI device, but its `_select_program_format` accepts only OpenQASM 3 or 2 and raises otherwise. This device advertises only `QIR_BASE_STRING` and `IQM_JSON`, so a `pennylane` extra would be dead on arrival until MQT Core's serializer registry covers the PennyLane path.
2. **`needs_calibration()` is not implemented** even though `QDMI_PROGRAM_FORMAT_CALIBRATION` is advertised. A client can trigger a calibration run but cannot learn that one is needed. No test pins this, since a test would freeze the gap rather than record it.

## Relationship to #189

#189 takes over `qiskit_to_iqm_json` and `MoveGate` from MQT Core. It stays blocked on v4: there is no `v4.0.0` tag, PyPI's latest `mqt-core` is 3.9.0, and 3.9.0 still ships both functions. Checked against the tag rather than the changelog — `serializers.py`, `register_program_serializer`, `_EXTRA_GATES`, `is_binary_program_format`, and the `mqt.core.qiskit.program_serializers` entry point group are absent from `v3.9.0` and present on `main`.

What #189 carried that 3.9.0 did release — the driver namespace move, the sampler and estimator removal, and the two `options=` call sites — is in this PR. #189 should rebase onto `main` after this merges and shrink to the v4-only half. 3.9.0 publishes wheels for every platform this repo builds, so none of #189's temporary MLIR CI scaffolding is needed here.

## Validation

- `uv lock` and `uv sync --group test` — `mqt-core` 3.9.0 resolves from a wheel, nothing builds from source
- `uv run python -W error::DeprecationWarning -c "import iqm.qdmi.qiskit"` — clean, which is the check that the `fomac` migration is complete
- `uv run pytest test/python` — 36 passed, 7 skipped; the skips are the live-hardware tests, which need `IQM_TOKEN`
- `uvx nox@latest -s lint` — full `prek` hook set, including `ty`
- CMake configure and build against the `v1.3.3` pin, then `ctest --test-dir …/test/unit` — 119/119 pass. The `test/integration/` failures are HTTP 401 from absent credentials.
- `uv run --script examples/qsci_h2.py --help` — builds the local package and parses the edited estimator call
- `uvx nox@latest -s docs` — `examples.md` and `python_package.md` execute cleanly, which covers the section added here. The build then fails in `qiskit.md`, whose first cell opens a live Resonance session and gets HTTP 401. That file is untouched by this PR and fails the same way on `main` without `IQM_TOKEN`, so the docs build is **not** fully verified locally. CI has the token and is the real check.

Not run: the live IQM and Resonance paths, which need credentials this machine does not have. The five new tests self-skip without them and first execute in CI, against `emerald:mock`.

## Checklist

- [x] Tests added for the new behavior
- [x] `CHANGELOG.md` updated
- [x] License headers and SPDX identifiers preserved
- [x] Lint and targeted tests pass

AI assistance: written by Claude Opus 5 via Claude Code, under my direction, and reviewed by me before submission.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,9 +60,9 @@ releases may include breaking changes.
 - ⬆️ Update MQT Core to 3.9.0, moving to its `mqt.core.qdmi.driver` namespace
   and its explicit sampler and estimator shot and precision defaults ([#195])
   ([**@marcelwa**])
-- ⬆️ Pin QDMI to the released v1.3.3 tag instead of a `develop` commit, and
-  publish the device target's stable ID and symbol prefix through QDMI's
-  `configure_qdmi_device_target` ([#195]) ([**@marcelwa**])
+- ⬆️ Update QDMI to v1.3.3, and publish the device target's stable ID and symbol
+  prefix through QDMI's `configure_qdmi_device_target` ([#195])
+  ([**@marcelwa**])
 - ⚡️ Reuse HTTP connections within each QDMI device session to reduce TCP/TLS
   setup during initialization and subsequent requests ([#163])
   ([**@burgholzer**])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ releases may include breaking changes.
 
 ### Added
 
+- ✅ Cover the device's queue length, a queued job's position, retrieval of an
+  existing job by ID, and calibration-format advertisement through MQT Core's
+  Python QDMI API ([#195]) ([**@marcelwa**])
 - 👷 Add an `IQM_QDMI_SANITIZERS` CMake option for building with
   AddressSanitizer, UndefinedBehaviorSanitizer, ThreadSanitizer, or
   MemorySanitizer, and run the C++ test suite under ASan and UBSan in CI
@@ -50,6 +53,12 @@ releases may include breaking changes.
 
 ### Changed
 
+- ⬆️ Update MQT Core to 3.9.0, moving to its `mqt.core.qdmi.driver` namespace
+  and its explicit sampler and estimator shot and precision defaults ([#195])
+  ([**@marcelwa**])
+- ⬆️ Pin QDMI to the released v1.3.3 tag instead of a `develop` commit, and
+  publish the device target's stable ID and symbol prefix through QDMI's
+  `configure_qdmi_device_target` ([#195]) ([**@marcelwa**])
 - ⚡️ Reuse HTTP connections within each QDMI device session to reduce TCP/TLS
   setup during initialization and subsequent requests ([#163])
   ([**@burgholzer**])
@@ -179,6 +188,7 @@ Compatible with QDMI `v1.3.0`.
 
 <!-- PR links -->
 
+[#195]: https://github.com/iqm-finland/QDMI-on-IQM/pull/195
 [#190]: https://github.com/iqm-finland/QDMI-on-IQM/pull/190
 [#188]: https://github.com/iqm-finland/QDMI-on-IQM/pull/188
 [#187]: https://github.com/iqm-finland/QDMI-on-IQM/pull/187
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ releases may include breaking changes.
 ### Added
 
 - ✅ Cover the device's queue length, a queued job's position, retrieval of an
-  existing job by ID, and calibration-format advertisement through MQT Core's
+  existing job by ID, and its advertised program formats through MQT Core's
   Python QDMI API ([#195]) ([**@marcelwa**])
 - 👷 Add an `IQM_QDMI_SANITIZERS` CMake option for building with
   AddressSanitizer, UndefinedBehaviorSanitizer, ThreadSanitizer, or
--- a/docs/python_package.md
+++ b/docs/python_package.md
@@ -225,19 +225,3 @@ print(job.check())
 ```
 
 A retrieved job cannot be resubmitted, and its parameters cannot be changed.
-
-### Triggering a Calibration Run
-
-Where the quantum computer supports it, the device advertises
-`ProgramFormat.CALIBRATION` among its supported program formats, and
-`submit_calibration_job` starts a run. A calibration run executes no circuit, so
-it takes no shot count, and the optional payload is a configuration for the run:
-
-```python
-job = device.submit_calibration_job()
-```
-
-```{warning}
-A calibration run is a real operation on a shared quantum computer. Trigger one
-only with the operator's agreement.
-```
--- a/docs/python_package.md
+++ b/docs/python_package.md
@@ -177,3 +177,67 @@ qc.measure_all()
 counts = sample(qc, shots=512, simulator=True)
 print("Counts:", counts)
 ```
+
+## Querying the Device Directly
+
+The Qiskit backend covers circuit execution, but a QDMI device also answers
+questions about itself. Open a session with MQT Core's driver to reach them.
+Constructing an {py:class}`~iqm.qdmi.qiskit.IQMBackend` registers the device
+under the stable ID {py:data}`~iqm.qdmi.IQM_QDMI_DEVICE_ID`, after which
+`open_device` resolves it:
+
+```python
+from mqt.core.qdmi.driver import open_device
+
+from iqm.qdmi import IQM_QDMI_DEVICE_ID
+
+device = open_device(IQM_QDMI_DEVICE_ID, token="…", custom2="emerald")
+
+print(device.status())
+print(device.supported_program_formats())
+```
+
+### Queue Length and Queue Position
+
+The device reports how busy the quantum computer is, so a client can decide
+whether to submit now or wait:
+
+```python
+waiting = device.queue_length()  # jobs waiting, excluding those executing
+```
+
+`queue_length()` returns `None` when the IQM API does not supply a trustworthy
+value. A queued job reports how many jobs are ahead of it; querying it refreshes
+the job's status first:
+
+```python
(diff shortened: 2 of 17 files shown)


test/python/test_device_capabilities.py
```diff
@@ -0,0 +1,134 @@
...
+# the License.
+#
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""Tests for device capabilities reached through MQT Core's QDMI client API.
+
+The device implements queue reporting, retrieval of an existing job by its ID,
+and calibration jobs. MQT Core 3.9.0 is the first release whose Python API
+reaches all three, so these tests cover the seam between the two packages
+rather than the device internals, which `test/unit/` already covers.
```
burgholzer: This oddly feels like an iteration artifact. At least the way it is worded.
This is not how one would have described this if this were genuine tests written for the IQM device, wouldn't it?

docs/python_package.md
```diff
@@ -177,3 +177,67 @@ qc.measure_all()
...
+it takes no shot count, and the optional payload is a configuration for the run:
+
+```python
+job = device.submit_calibration_job()
+```
+
+```{warning}
+A calibration run is a real operation on a shared quantum computer. Trigger one
+only with the operator's agreement.
+```
```
burgholzer: This doesn't make too much sense to me. Only very select few people will be allowed to trigger calibration.
Also, IQM's calibration requires a configuration for the payload.
I am not sure we should be describing/advertising this here.

burgholzer (changes requested):
This largely looks good, thanks @marcelwa 🙏🏼 

I have one more general request for changes:
`https://github.com/iqm-finland/QDMI-on-IQM/blob/agent/mqt-core-3.9-qdmi-1.3.3/python/iqm/qdmi/_backends.py` can take advantage of the stable device ID lookup for DDSIM (so can likely a few other places). I also think the `build_sampler` and `build_estimator` signatures could be simplified to only return the primitive. The backend is available from the primitives in any case, which reduces the akward `tuple` return.

docs/python_package.md
marcelwa: Totally agree on the warning artifact. Do you want to drop the entire configuration job description?

test/python/test_device_capabilities.py
marcelwa: 🤖 *AI text below* 🤖

You're right, and it was. That paragraph described the change that introduced the file rather than the tests, and naming the release that first exposed the APIs would have aged into noise. Rewritten in 4787401 to say what the tests drive and how they differ from the C API tests:

```
Tests for the device's queue, job-retrieval, and calibration capabilities.

These drive the device through MQT Core's Python QDMI client, the way an
application reaches it. `test/unit/` covers the same properties at the C API
level, against a stubbed server.
```

Same trimming applied to the format test's docstring, which had picked up a similar aside.

docs/python_package.md
burgholzer: Yeah, I think so 👍🏼

marcelwa: 🤖 *AI text below* 🤖

Dropped in 2f9859a — the whole `Triggering a Calibration Run` subsection is gone.

Worth recording why it was doubly wrong: beyond the access point you raise, the snippet could not have worked. `submit_calibration_job()` with no payload leaves the job's program unset, and `src/iqm_device.cpp:1313` rejects that with `QDMI_ERROR_INVALIDARGUMENT` before anything reaches the API.

The changelog entry no longer singles calibration out either. The test suite still pins the advertised program-format set with calibration included, since that is a device contract rather than something to point users at — shout if you would rather that went too.

marcelwa:
The flakey test should be resolved once #196 lands.

python/iqm/qdmi/_backends.py
```diff
@@ -45,18 +40,18 @@ def build_sampler(
...
 
     Returns:
-        A tuple of the resolved backend and a Sampler primitive bound to it.
+        A Sampler primitive bound to the resolved backend, which it exposes as
+        its `backend` property.
     """
     if simulator:
-        backend = QDMIProvider().get_backend(_SIMULATOR_DEVICE)
-        return backend, QDMISampler(backend)
+        return QDMISampler(QDMIBackend.from_device_id(_SIMULATOR_DEVICE_ID))
```
burgholzer:
```suggestion
        return QDMIBackend.from_device_id(_SIMULATOR_DEVICE_ID).sampler()
```
Should just work, right?

python/iqm/qdmi/_backends.py
```diff
@@ -66,15 +61,15 @@ def build_estimator(
...
 
     Returns:
-        A tuple of the resolved backend and an Estimator primitive bound to it.
+        An Estimator primitive bound to the resolved backend, which it exposes
+        as its `backend` property.
     """
     if simulator:
-        backend = QDMIProvider().get_backend(_SIMULATOR_DEVICE)
-        return backend, QDMIEstimator(backend)
+        return QDMIEstimator(QDMIBackend.from_device_id(_SIMULATOR_DEVICE_ID))
```
burgholzer:
```suggestion
        return QDMIBackend.from_device_id(_SIMULATOR_DEVICE_ID).estimator()
```
Same here.

CHANGELOG.md
```diff
@@ -54,6 +57,12 @@ releases may include breaking changes.
 
 ### Changed
 
+- ⬆️ Update MQT Core to 3.9.0, moving to its `mqt.core.qdmi.driver` namespace
+  and its explicit sampler and estimator shot and precision defaults ([#195])
+  ([**@marcelwa**])
+- ⬆️ Pin QDMI to the released v1.3.3 tag instead of a `develop` commit, and
```
```suggestion
- ⬆️ Update QDMI to v1.3.3, and
```
Released versions always pin a released QdMI version. So the previous entry made no sense.
This will likely trigger a reformat.

burgholzer (changes requested):
Three small corrections before this can go in. After that, it's getting time to get together the next release. So let's collect the issues/PRs that should still make it into that release.

marcelwa:
@burgholzer since your suggestions moved `QDMISampler` and `QDMIEstmator` into the `TYPE_CHECKING` block, and since `CHANGELOG.md` ended up being reformatted, I did not directly apply your suggestions, but created two new commits that attribute you as co-author.

burgholzer (approved):
LGTM now 👍🏼

PR #196, opened 2026-08-21
✅ Stop the queue-position test failing when a job outruns the queue

marcelwa:
🤖 *AI text below* 🤖

`QDMIIntegrationTest.QueuePropertiesOnGarnetMock` fails intermittently, and the reason is a race the test cannot win. It submits four 64-shot mock jobs and polls the last one for a queue position. A status request that hits HTTP 429 absorbs a `Retry-After` wait of tens of seconds, while a 64-shot mock job finishes in well under a second, so by the time the request returns the job has left the queue and the property correctly answers `QDMI_ERROR_BADSTATE`. The test breaks out of its poll loop on that code and then hard-fails on `EXPECT_TRUE(observed_queue_position.has_value())`. In the garnet job of [run 31381541803](https://github.com/iqm-finland/QDMI-on-IQM/actions/runs/31381541803) there were six 429s with a 30-second `Retry-After`, one of them on a job-status GET, and the test failed all three `ctest --repeat until-pass:3` attempts.

This supersedes #184, which tried to fix it in the device by answering the first query from the submission response. That was the wrong place. QDMI states in `include/qdmi/constants.h` that querying `QDMI_DEVICE_JOB_PROPERTY_QUEUEPOSITION` *must* refresh the status and queue position, and *must* return `QDMI_ERROR_BADSTATE` when the refreshed job is not queued. In the failing scenario `BADSTATE` is the specified answer, so the device is already right and the assertion is what is wrong. Thanks to @burgholzer for the pushback.

The test now treats a job that outran the queue as the legitimate outcome it is, and keeps asserting what does hold — that the property only ever answers with a position, `QDMI_ERROR_NOTSUPPORTED`, or `QDMI_ERROR_BADSTATE`. It also polls half as often over the same window. This is the tightest polling loop in the suite, since `IQM_QDMI_device_job_wait` already backs off exponentially from one second, so it is a plausible contributor to the rate limiting — though six 429s in one job does not isolate the source, so treat that as a guess rather than a diagnosis.

Two unit tests take over what the integration test can no longer guarantee. `SubmissionQueuePositionDoesNotSkipTheRefresh` and `CalibrationSubmissionQueuePositionDoesNotSkipTheRefresh` pin the property to the server on both submit paths, even when the submission response already carried a position — they fail against #184's implementation, which is the point of adding them. Both assert on the number of requests rather than on a value alone, because `HttpStub` never verifies that a queued response was consumed.

The one production change left is a tidy: both submit paths carried the same block re-parsing `queue_position` for their log line. That now lives in one helper and reads the value as an unsigned number the way `IQM_QDMI_device_job_check` does, so a negative position is no longer logged as if it were real. An earlier revision also adopted the submission response's native status; @burgholzer's review showed that no public behavior observed it and that it could turn an unknown native status into a logged error on a successful submission, so it is gone.

No `CHANGELOG.md` entry: a flaky test and a log-format tidy are not user-facing behavior. Flagging that as a judgment call rather than leaving it silent.

Full unit suite green at 124 tests on top of current `main`. `uvx nox -s lint` clean, and `clang-tidy` clean on all three changed files. The integration test needs live credentials and `garnet:mock`, so CI exercises it — note that a single green run proves nothing about a flake here; what changed is that the `BADSTATE` path can no longer fail the test.

### DIFF
--- a/src/iqm_device.cpp
+++ b/src/iqm_device.cpp
@@ -916,6 +916,32 @@ int Set_job_status(IQM_QDMI_Device_Job job, const std::string &native_status) {
   }
   return QDMI_SUCCESS;
 }
+
+/**
+ * @brief Adopt the native status that a submission response reports, as long as
+ *        it says the job is queued.
+ * @details A submission response carries the job's native status, which is a
+ *          more precise observation than the generic submitted status. Only a
+ *          queued observation is adopted: a terminal status would make @ref
+ *          IQM_QDMI_device_job_done consider the job finished and keep @ref
+ *          IQM_QDMI_device_job_wait from ever polling for it.
+ * @param job The job that was just submitted.
+ * @param response The parsed submission response.
+ */
+void Adopt_submission_status(IQM_QDMI_Device_Job job,
+                             const nlohmann::json &response) {
+  const auto native_status = response.find("status");
+  if (native_status == response.end() || !native_status->is_string()) {
+    return;
+  }
+  // Run the observation through the regular mapping rather than repeating the
+  // native status names here.
+  const auto submitted_status = job->status_;
+  if (Set_job_status(job, native_status->get<std::string>()) != QDMI_SUCCESS ||
+      job->status_ != QDMI_JOB_STATUS_QUEUED) {
+    job->status_ = submitted_status;
+  }
+}
 } // namespace
 
 int IQM_QDMI_device_session_retrieve_device_job_by_id(
@@ -1235,6 +1261,7 @@ int IQM_QDMI_device_job_submit_circuit(IQM_QDMI_Device_Job job) {
 
   job->job_id_ = job_submission_json_response.at("id").get<std::string>();
   job->status_ = QDMI_JOB_STATUS_SUBMITTED;
+  Adopt_submission_status(job, job_submission_json_response);
 
   // Log queue position if available
   std::string log_message = "Submitted job with ID: " + job->job_id_;
@@ -1281,6 +1308,7 @@ int IQM_QDMI_device_job_submit_calibration(IQM_QDMI_Device_Job job) {
 
   job->job_id_ = job_submission_json_response.at("id").get<std::string>();
   job->status_ = QDMI_JOB_STATUS_SUBMITTED;
+  Adopt_submission_status(job, job_submission_json_response);
 
   // Log queue position if available
   std::string log_message =
--- a/src/iqm_device.cpp
+++ b/src/iqm_device.cpp
@@ -931,32 +931,6 @@ Submission_queue_position(const nlohmann::json &response) {
   }
   return position->get<size_t>();
 }
-
-/**
- * @brief Adopt the native status that a submission response reports, as long as
- *        it says the job is queued.
- * @details A submission response carries the job's native status, which is a
- *          more precise observation than the generic submitted status. Only a
- *          queued observation is adopted: a terminal status would make @ref
- *          IQM_QDMI_device_job_done consider the job finished and keep @ref
- *          IQM_QDMI_device_job_wait from ever polling for it.
- * @param job The job that was just submitted.
- * @param response The parsed submission response.
- */
-void Adopt_submission_status(IQM_QDMI_Device_Job job,
-                             const nlohmann::json &response) {
-  const auto native_status = response.find("status");
-  if (native_status == response.end() || !native_status->is_string()) {
-    return;
-  }
-  // Run the observation through the regular mapping rather than repeating the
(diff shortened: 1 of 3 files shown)


marcelwa:
🤖 *AI text below* 🤖

One thing in `Adopt_submission_status` is worth calling out, because it reads like defensive coding and is not.

The helper adopts the submission response's native status only when it maps to `QDMI_JOB_STATUS_QUEUED`, and restores the submitted status otherwise. That restriction is load-bearing. `Set_job_status` maps `"ready"` and `"completed"` to `QDMI_JOB_STATUS_DONE`, `"aborted"` and `"cancelled"` to `QDMI_JOB_STATUS_CANCELED`, and `"failed"` to `QDMI_JOB_STATUS_FAILED`. `IQM_QDMI_device_job_done` reads `job->status_` directly, and `IQM_QDMI_device_job_check` returns early on exactly those three. So a submission response carrying a terminal status would leave the job believing it had already finished, and `IQM_QDMI_device_job_wait` would return `QDMI_SUCCESS` immediately without ever issuing a status request — for a job that in reality is still running remotely.

Dropping the guard and adopting whatever the response says would look like a simplification and would pass every test that existed before this PR. `SubmissionStatusIsAdoptedOnlyWhileQueued` is there to stop that: it submits with `{"status": "failed"}` and asserts that the following `IQM_QDMI_device_job_check` still issues a request and still observes the real status.

Whether the IQM Server ever answers a submission with a terminal status is a separate question, and I have not seen it do so. The guard does not depend on that: the mapping makes it reachable, and the failure mode if it is ever reached is silent rather than loud.

burgholzer:
I looked through this and got a bit confused myself about the `Adopt_submission_status` function. I think the review below pins this down more specifically.
I'd be in favor of dropping that function.

🤖 *AI text below* 🤖

**[P2] Remove the unobservable submission-status adoption**

[`Adopt_submission_status`](https://github.com/iqm-finland/QDMI-on-IQM/blob/4b1a3f3986b0bc925c7785381dfa1f217e56afd9/src/iqm_device.cpp#L946-L958) does not appear to establish any observable public behavior. Job-status checks refresh from the server, `QDMI_DEVICE_JOB_PROPERTY_QUEUEPOSITION` must refresh by contract, and the remaining entry points treat `QDMI_JOB_STATUS_SUBMITTED` and `QDMI_JOB_STATUS_QUEUED` identically.

The helper nevertheless calls `Set_job_status` for every string-valued submission status. A future native status that this client does not know therefore produces `LOG_ERROR("Unknown job status: ...")` before the helper silently restores `QDMI_JOB_STATUS_SUBMITTED`, turning an otherwise successful submission into a false operational error.

The added tests do not pin the adoption itself: they queue a subsequent status GET and assert its refreshed result, so they would still pass if the helper and its two call sites were removed. I recommend removing the adoption path and its terminal-status guard test. The circuit and calibration tests can remain, renamed to describe the useful behavior they actually cover: submission metadata must not skip the mandatory queue-position refresh.

The integration-test relaxation itself looks correct and keeps the QDMI ownership boundary intact; this production path is unrelated scope that adds risk without supporting that fix.

Local validation at `4b1a3f3`: Release build passed, all 122 unit tests passed, and `uvx nox -s lint` passed. I did not run live IQM integration tests locally; hosted CI is green.

burgholzer (changes requested):
Just so this is formally marked as a request for changes 😌

marcelwa:
🤖 *AI text below* 🤖

Removed in 23acdda — `Adopt_submission_status` and both call sites are gone, along with `SubmissionStatusIsAdoptedOnlyWhileQueued`.

You are right that nothing observed the adoption, and the unknown-status argument settles it on its own: running every string-valued submission status through `Set_job_status` means a native status this client does not know logs an error and turns a successful submission into a false fault, in exchange for state no caller can read. My earlier comment defended the terminal-status guard on its own terms and never asked whether the thing being guarded was worth having.

The two queue-position tests stay. `CalibrationSubmissionAdoptsItsQueuedStatus` is renamed to `CalibrationSubmissionQueuePositionDoesNotSkipTheRefresh` to match its sibling and describe what it covers. Both still submit a response carrying a position and assert that the query issues its own GET and returns the server's value — they fail against #184.

What remains in `src/iqm_device.cpp` is the `Submission_queue_position` helper for the log line, which deduplicates the block both submit paths carried and reads the value as unsigned.

Also merged current `main` in. 124 unit tests green, `uvx nox -s lint` clean.

burgholzer (approved):
Yeah, this looks good to me now. Let's get this in 👍🏼

PR #199, opened 2026-08-21
🐛 Canceling a job that already finished reports a denied permission

### LINKED ISSUE #197: 🐛 JobCancellation integration test fails when the cancel request is rate limited
🤖 *AI text below* 🤖

## System and Environment Information

- Operating system and version: `ubuntu-24.04` GitHub Actions runner
- IQM QDMI Device version: `main` at 47be60d, plus the branch of #196
- Language versions: C++20
- Compiler or build tool versions: the `🇨 Sanitizers / 🧪 ASan + UBSan` job (Debug, `-fsanitize=address,undefined`)
- Any relevant dependencies and their versions: live `resonance.iqm.tech`, `emerald:mock`

## Bug Description

`QDMIIntegrationTest.JobCancellation` fails intermittently when the cancellation request is rate limited. The job finishes while the client waits out the `Retry-After`, and cancelling a finished job is then rejected, so both assertions in the test fail.

From the sanitizer job of [run 32476694628](https://github.com/iqm-finland/QDMI-on-IQM/actions/runs/32476694628):

```text
INFO  Submitted job with ID: 01a0240e-… (queue position: 2)
INFO  Canceling job with ID: 01a0240e-…
DEBUG Request to URL '…/jobs/01a0240e-…/cancel' hit HTTP 429 rate limiting; retrying after 30 second(s)
ERROR Request to URL '…/jobs/01a0240e-…/cancel' failed with HTTP 403 (Client Error)
ERROR Error: [illegal_job_status] Illegal job status
DEBUG Job status: 4 (native status: completed)
```

```text
test_iqm_device_integration.cpp:899: Failure
  IQM_QDMI_device_job_cancel(job)  Which is: -8
  QDMI_SUCCESS                     Which is: 0
test_iqm_device_integration.cpp:902: Failure
  status                     Which is: 4
  QDMI_JOB_STATUS_CANCELED   Which is: 5
```

The test took 32.5 seconds, essentially all of it the rate-limit wait.

The device behaved correctly throughout. A job that has already completed cannot be cancelled, and `IQM_QDMI_device_job_cancel` deliberately leaves the status untouched when the request fails. What is wrong is the test: it asserts that a mock job submitted moments earlier is still cancellable, which a shared and rate-limited backend cannot promise. A 64-shot mock job finishes in under a second, while a single 429 costs 30 seconds.

This is the same failure mode as the `QueuePropertiesOnGarnetMock` flake that #196 addresses — a test asserting on a transient state of a live service across a request that can outlive that state — but in a different test, so filing it separately rather than widening that PR.

## Steps to Reproduce

1. Run the integration suite against `resonance.iqm.tech` with `emerald:mock` while the endpoint is rate limiting.
2. `ctest -R JobCancellation`.
3. Observe the cancel POST take an HTTP 429 with a 30-second `Retry-After`, and the retry return HTTP 403 `illegal_job_status` because the job completed in the meantime.

Reproducing on demand needs the runner to actually be rate limited, which is what makes it intermittent. The mechanism is fully visible in the log above.

## Suggested fix

Accept the outcome instead of asserting against it: treat a cancellation rejected because the job already reached a terminal state as a valid result, and assert `QDMI_JOB_STATUS_CANCELED` only when the cancellation was actually accepted. The cancellation contract itself is worth pinning in `test/unit` against `HttpStub`, where the timing is deterministic.

marcelwa:
🤖 *AI text below* 🤖

Closes #197.

Canceling a job that the IQM Server has already finished returned `QDMI_ERROR_PERMISSIONDENIED`. QDMI defines that code for this function as the device not allowing the [device job interface](https://munich-quantum-software-stack.github.io/QDMI/) for the current session, so a client that cancels a job it submitted moments earlier and gets that answer would reasonably conclude its token is wrong — when in fact its job had simply completed. The cancel path's own `catch (const iqm::ClientAuthenticationError &)` returns the same code, so the two were indistinguishable.

The device also contradicted itself. `IQM_QDMI_device_job_cancel` already returns `QDMI_ERROR_INVALIDARGUMENT` when it knows locally that the job is `QDMI_JOB_STATUS_DONE`, which is what QDMI documents for a job that can no longer be canceled and what `JobCycleCornerCases` already asserts. The same fact produced a different answer depending only on whether this device happened to have refreshed its status.

`iqm::http::Handle_response` maps every HTTP 403 to `QDMI_ERROR_PERMISSIONDENIED` regardless of the body. That is right in general — 403 does mean permission denied elsewhere — so the fix stays inside the cancel path rather than changing the shared mapping. The server distinguishes the two cases itself with an `illegal_job_status` error code, which `Handle_response` already parses and logs but does not act on. The new check keys on that error code rather than on the status code, so it keeps working if the server ever answers with a different one, and it reads both shapes the code arrives in: at the root of the response, and inside an `errors` array.

The job status is deliberately left alone. The refusal establishes that the job is terminal but not which terminal status it reached — `completed`, `failed`, and `cancelled` are all possible — so the next `IQM_QDMI_device_job_check` is what observes it.

## The test

`QDMIIntegrationTest.JobCancellation` required a mock job to still be cancellable moments after submission. That is not something a shared backend can promise: a 64-shot job finishes in well under a second, and a rate-limited cancel request waits out a `Retry-After` of tens of seconds before it arrives. It failed exactly that way in the sanitizer job of [run 32476694628](https://github.com/iqm-finland/QDMI-on-IQM/actions/runs/32476694628), taking 32.5 seconds, essentially all of it the block.

It now asserts against whichever outcome occurred: a cancellation that succeeds must leave the job canceled, and one that is refused must be refused with `QDMI_ERROR_INVALIDARGUMENT` and leave the job in some terminal status. Both branches carry a real assertion, so this loosens what the test demands of the backend without loosening what it demands of the device.

There is no second race on the success path, which is why no polling is needed: a successful cancellation sets the status to `QDMI_JOB_STATUS_CANCELED` locally, and `IQM_QDMI_device_job_check` returns early for terminal statuses without issuing a request.

## Testing

`CancelingAJobThatAlreadyFinishedIsAnInvalidArgument` drives both response shapes against the HTTP stub and asserts the job stays observable afterwards, with a request-count delta rather than a value check alone. It fails against the previous mapping, returning `-8` where `-7` is expected, which is the point of adding it. The neighboring `JobCancelPreservesStatusOnServerError` and the existing 403-without-an-error-code case are untouched, and the latter is what keeps this mapping from widening into genuine authorization failures. A third test pins that a refusal whose body is not JSON at all keeps the mapping the status code alone would have produced.

Full unit suite green at 121 tests, `uvx nox -s lint` clean, and `clang-tidy` clean on all three changed files.

## Note on sequencing

This is independent of #196 and reviewable on its own — different function, different test — though both touch `test/integration/test_iqm_device_integration.cpp` in well-separated places, so whichever lands second may want a trivial rebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ releases may include breaking changes.
 
 ### Fixed
 
+- 🐛 Report a cancellation that the server refuses because the job already
+  finished as `QDMI_ERROR_INVALIDARGUMENT` rather than
+  `QDMI_ERROR_PERMISSIONDENIED`, which said the session was not allowed to use
+  the job interface ([#199]) ([**@marcelwa**])
 - 🐛 Report the status the quantum computer is actually in, instead of pinning a
   session to busy from its first job submission onwards ([#190])
   ([**@marcelwa**])
@@ -179,6 +183,7 @@ Compatible with QDMI `v1.3.0`.
 
 <!-- PR links -->
 
+[#199]: https://github.com/iqm-finland/QDMI-on-IQM/pull/199
 [#190]: https://github.com/iqm-finland/QDMI-on-IQM/pull/190
 [#188]: https://github.com/iqm-finland/QDMI-on-IQM/pull/188
 [#187]: https://github.com/iqm-finland/QDMI-on-IQM/pull/187
--- a/src/iqm_device.cpp
+++ b/src/iqm_device.cpp
@@ -34,6 +34,7 @@
 #include <cassert>
 #include <chrono>
 #include <cpr/connection_pool.h>
+#include <cpr/response.h>
 #include <cstdint>
 #include <cstdlib>
 #include <cstring>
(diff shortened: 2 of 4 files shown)


burgholzer (approved):
LGTM 👍🏼

PR #202, opened 2026-08-21
✨ Let callers choose the Slurm partition and node count for offloaded jobs

marcelwa:
🤖 *AI text below* 🤖

`iqm.qdmi.offloader`'s `sample()` and `estimate()` hardcoded `--partition=quantum` and `--nodes=1` on the `srun` command line, so a site whose quantum partition carries a different name could not use the offloader at all. Both now take `partition` and `nodes` keyword arguments, threaded through the way `licenses` already is. `partition` resolves in the order explicit argument, then `IQM_SLURM_PARTITION`, then `quantum` — the precedence the SPANK plugin's `IQM_BASE_URL`/`IQM_QC_ID`/`IQM_QC_ALIAS` variables already follow. `nodes` gets no environment fallback: a node count is a per-job resource request rather than a site-wide constant. Nothing changes for a cluster already using `quantum`.

`--ntasks=1` is pinned beside `--nodes`. Without it `srun --nodes=N` defaults to one task per node, so any node count above one would launch `iqm-sampler`/`iqm-estimator` once per node — each copy submitting its own hardware job and writing its own base64 payload into the one captured stdout, where `_decode_payload` dies with a bare `binascii.Error: Incorrect padding`. `nodes` therefore sizes the allocation for a partition that demands more than one node; it does not parallelize the workload.

`docs/python_package.md` spells out the resolution order and gives `nodes` its own section. The administrator guide now names both places a renamed partition has to reach: `IQM_SLURM_PARTITION` for the offloader, and the SPANK plugin's own `partitions=` option — miss that one and the plugin skips those jobs and injects no `IQM_BASE_URL`, `IQM_QC_ID`, or `IQM_QC_ALIAS`, with only a debug-level log line to show for it. `IQM_SLURM_PARTITION` is listed in `.env.example` too.

Tests cover the resolution order for both entry points, the default and an explicit node count, and `--ntasks=1`, asserting by index rather than by membership since an option landing after the worker becomes the worker's own argument. `uvx nox@latest -s tests-3.14` reports 58 passed, 7 skipped (the skips are the live IQM backend tests, which need credentials); `uvx nox@latest -s lint` passes. No live integration workflow was run. `main` is merged in; the v1.4.0 cut moved this branch's changelog entry under `## [Unreleased]`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -54,3 +54,8 @@
 # Directory for job files (circuit QPY, operator PKL files)
 # Default: ~/.qdmi_jobs (user's home directory under .qdmi_jobs)
 #IQM_JOBS_DIR=
+
+# Slurm partition the offloader submits its srun jobs to
+# Default: quantum
+# Example: IQM_SLURM_PARTITION=qc-nodes
+#IQM_SLURM_PARTITION=
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,9 @@ releases may include breaking changes.
 ### Added
 
 - ✨ Add `partition` and `nodes` keyword arguments to `iqm.qdmi.offloader`'s
-  `sample`/`estimate`, resolving the partition from `IQM_SLURM_PARTITION` when
-  it is not passed, so a site whose quantum partition is not named `quantum` can
-  use the offloader ([#202]) ([**@marcelwa**])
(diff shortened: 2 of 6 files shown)


burgholzer (approved):
LGTM 👍🏻

PR #204, opened 2026-08-21
🐛 Stop a malformed calibration set from crashing or misdescribing the device

marcelwa:
🤖 *AI text below* 🤖

`Process_calibration_metrics` treated the dynamic architecture's `loci` arity as a programmer invariant and checked it with `assert()`. The project's default `CMAKE_BUILD_TYPE` is `Release`, so `NDEBUG` removes those checks from the configuration that actually ships, and the function goes on to index the locus and dereference what it finds. Server data was deciding whether the driver stayed inside its own memory.

Three shapes were reachable, all confirmed against an `-DNDEBUG` ASan/UBSan build: a `cz` entry with a one-element locus read past the vector (heap-buffer-overflow), and a `prx` or `measure` entry with an empty locus bound a reference to a null pointer and then dereferenced it for the site name. In a plain Release build the empty-locus case is a segfault during session initialization.

The same function converted `characterization.model.<qubit>.t1_time` and `.t2_time` with `static_cast<uint64_t>(value * 1e6)`. A value of `2e111`, or any negative one, is a `float-cast-overflow`. These values parse fine and have the right JSON type — they are the wrong magnitude — so the JSON access hardening in #188 does not reach them.

## The property interface trusted the same data

Guarding the metrics loop alone left the shape of `loci` unchecked where it decides how a caller reads a result. `QDMI_OPERATION_PROPERTY_QUBITSNUM` reports `available_sites_for_op.front().size()` — the arity of the *first* locus — while `QDMI_OPERATION_PROPERTY_SITES` flattens every locus into one list regardless of size. A `cz` reporting `[["QB1", "QB2"], ["QB1"]]` therefore advertises an arity of 2 over a list of 3 sites, and a caller slicing that list back into loci reads one element past the buffer it was handed. A gate reporting `"loci": []` has no first locus at all: the `assert` guarding `front()` is compiled out in Release, so querying its arity segfaults after initialization has already returned `QDMI_SUCCESS`.

## Validating the shape where it enters

Loci are now validated while the dynamic quantum architecture is parsed, and an operation is registered only once they hold the same non-zero number of sites. A gate whose loci are absent, site-less, or of differing sizes is dropped and named in the log, so the property interface never sees a gate it cannot describe. Because every registered gate then has a uniform arity, the arity check runs once per gate instead of once per locus, and the property interface rejects an empty site list rather than dereferencing past it.

A coherence time that is negative, not finite, larger than a `uint64_t` can hold, or below one microsecond is dropped and the metric named. That last case truncated to `0`, which `QDMI_SITE_PROPERTY_T1` already uses to mean "not reported", so an unusably small value was indistinguishable from an absent one.

Everything else in the calibration set is still processed, so one malformed entry no longer costs the whole session — the same degrade-rather-than-fail behavior this driver already applies to an absent optional endpoint. `assert` stays for invariants the server cannot influence.

Six regression tests drive the payloads through the existing `HttpStub`. Each asserts that initialization still succeeds, that the skipped entry appears in the log, and that the well-formed entries beside it keep their fidelity or coherence time; the two covering the property interface also assert that the malformed gate is absent and that a surviving gate's arity and site count stay consistent. Against the previous code the empty-locus case crashes with SIGSEGV, the one-element `cz` case ends in `QDMI_ERROR_FATAL`, and the mixed-arity case hands back three sites under an advertised arity of two. The ASan/UBSan job from #173 covers them a second way.

## Testing

Full unit suite green at 130 tests on this branch. `clang-tidy` clean on every changed file. Live integration tests were not run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,10 +33,8 @@ releases may include breaking changes.
 ### Fixed
 
 - 🐛 Survive a calibration set whose gate loci or coherence times have an
-  unexpected shape: drop a gate whose loci are absent, site-less, or of
-  differing sizes rather than advertising more sites than its reported arity
-  accounts for, and ignore a coherence time that no site property can represent
-  ([#204]) ([**@marcelwa**])
+  unexpected shape, dropping the affected gate or metric instead of crashing or
+  misreporting it ([#204]) ([**@marcelwa**])
 - 🐛 Report a cancellation that the server refuses because the job already
   finished as `QDMI_ERROR_INVALIDARGUMENT` rather than
   `QDMI_ERROR_PERMISSIONDENIED`, which said the session was not allowed to use
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,8 +33,10 @@ releases may include breaking changes.
 ### Fixed
 
 - 🐛 Survive a calibration set whose gate loci or coherence times have an
-  unexpected shape, instead of reading past the reported sites or converting an
-  out-of-range value ([#204]) ([**@marcelwa**])
+  unexpected shape: drop a gate whose loci are absent, site-less, or of
+  differing sizes rather than advertising more sites than its reported arity
+  accounts for, and ignore a coherence time that no site property can represent
+  ([#204]) ([**@marcelwa**])
 - 🐛 Report a cancellation that the server refuses because the job already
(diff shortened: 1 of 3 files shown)


burgholzer (approved):
It's a bit much on the defensive side for my taste, but overall this LGTM! So let's get it in! 👍🏻

PR #205, opened 2026-08-21
🐛 Fix token expiry beyond 2038 and keep raw response bodies out of default logs

marcelwa:
🤖 *AI text below* 🤖

Four small, independent fixes, one commit each so any of them can be dropped without disturbing the rest — with one follow-up commit refining the log-level change, which travels with it. Two are bugs a user can hit — token expiry arithmetic and what a failed request writes to the log; the other two are a hash-quality improvement and the removal of a duplicated function, neither of which changes an answer.

**This PR originally carried a fifth change** — moving the log level to `IQM_LOG_LEVEL` — which pulled in seven workflow files, `pyproject.toml`, the SPANK plugin, and three documents. That was most of the diff and none of the substance, so it now lives in #206 and this PR is back to what it says on the tin.

## 🐛 JWT expiry arithmetic in 64-bit time

`TokenManager::time_left_seconds` read `exp` into an `int` and compared it against `static_cast<int>(std::time(nullptr))`. Both truncate in January 2038, after which a valid token reads as expired and authentication fails outright. It now returns `int64_t`, parses `exp` as `int64_t`, and compares against an `int64_t` clock; `REFRESH_MARGIN_SECONDS` follows so the comparison stays homogeneous.

`TimeLeftSecondsHandlesExpiryBeyond2038` pins a token expiring 2100-01-01 and brackets the result between two `std::time` reads. It fails against the previous behavior.

## 🩹 Site-pair hash depends on the order of its elements

`Pair_hash` combined the two site hashes with `hash1 ^ hash2`, which is symmetric, so `(a, b)` and `(b, a)` landed in the same bucket of `operations_two_qubit_fidelity_map_`. `operator==` kept lookups *correct*, so this was never a wrong-answer bug — but the map is deliberately queried in both orientations, which makes every two-qubit entry a guaranteed collision rather than an unlucky one. Replaced with a golden-ratio mix sized to the hash width — the 32-bit constant leaves the top half of a 64-bit `size_t` under-mixed, which would undercut the only thing the change is for.

**Stated plainly: the added test is a guard, not a regression test.** `TwoQubitFidelityIsFoundInEitherSiteOrder` passes before and after, because the old behavior was slow rather than wrong. It pins that both orders keep resolving.

## ♻️ Drop `Get_optional`, a verbatim copy of `Get`

The two functions had identical bodies. The `ERROR_LOG_POLICY` parameter that would justify the split is never consumed inside either one — callers apply it when they invoke `Handle_response`, not when they fetch. Deleted along with its declaration.

Three call sites, all updated: the CoCoS health probe and `Probe_quantum_computer` in `src/iqm_device.cpp`, plus one in `test/unit/test_http_client.cpp`. Both production callers pass `LOG_AS_DEBUG` to `Handle_response`, so the rationale holds for each. The stale `Get`/`Get_optional`/`Post` comment in `http_stub.hpp` is fixed too.

Not an API break: `include/` is not installed — only the generated prefixed QDMI headers ship — so no changelog entry.

## 🔒️ Keep raw HTTP response bodies out of ERROR-level logs

For an error response carrying no structured error field, `Handle_response` fell back to logging the whole body at ERROR, which is on by default. A server that echoes request contents therefore surfaced them without anyone opting in, and `AGENTS.md` requires never exposing token contents.

**Demotion to DEBUG rather than a redaction pass**, deliberately: redaction would be a heuristic over a body shape nobody specifies, where every miss is a silent leak, whereas the level boundary is exact and DEBUG already dumps raw POST payloads.

The default level *is* ERROR, though, so demoting the body alone would have made an opaque upstream failure — a proxy's HTML, a gateway's plain text — invisible without setting an environment variable and reproducing it. The ERROR line therefore reports the body's size and content type alongside the status. Neither carries request content, and together they are usually enough to tell an upstream failure from one the device produced.

`RawResponseBodyStaysOutOfErrorLevelLogs` asserts the secret stays out of ERROR while the size and type appear; `UntypedResponseBodyIsReportedWithoutAContentType` and `RawResponseBodyReachesDebugLevelLogs` cover the missing-header path and the body still reaching DEBUG.

## Testing

Full unit suite green at 129 tests on this branch. `uvx nox -s lint` clean. `clang-tidy` clean on every changed file. Live integration tests were not run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,8 @@ releases may include breaking changes.
 
 - 🔒️ Log raw HTTP error-response bodies at `DEBUG` instead of `ERROR`, so a
   response that echoes request contents no longer surfaces at the default log
-  level ([#205]) ([**@marcelwa**])
+  level, while still reporting the body's size and content type at `ERROR` so an
+  opaque upstream failure stays diagnosable ([#205]) ([**@marcelwa**])
 - 🐛 Keep authentication working past January 2038 by reading JWT expiry times
   as 64-bit values ([#205]) ([**@marcelwa**])
 - 🐛 Report a cancellation that the server refuses because the job already
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,10 +32,9 @@ releases may include breaking changes.
 
 ### Fixed
 
-- 🔒️ Log raw HTTP error-response bodies at `DEBUG` instead of `ERROR`, so a
-  response that echoes request contents no longer surfaces at the default log
-  level, while still reporting the body's size and content type at `ERROR` so an
-  opaque upstream failure stays diagnosable ([#205]) ([**@marcelwa**])
+- 🔒️ Log raw HTTP error-response bodies at `DEBUG` instead of `ERROR`, keeping
+  only their size and content type at the default log level ([#205])
+  ([**@marcelwa**])
 - 🐛 Keep authentication working past January 2038 by reading JWT expiry times
   as 64-bit values ([#205]) ([**@marcelwa**])
 - 🐛 Report a cancellation that the server refuses because the job already
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,11 @@ releases may include breaking changes.
 
 ### Fixed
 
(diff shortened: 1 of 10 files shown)


burgholzer (approved):
Nice. This can happily go in once the conflicts are resolved!

PR #206, opened 2026-08-21
✨ Accept `IQM_LOG_LEVEL` as the log level environment variable

marcelwa:
🤖 *AI text below* 🤖

Split out of #205, which grew too large to review in one pass. That PR keeps the four small correctness fixes; this one carries the environment-variable change, which is where all of its breadth came from.

`IQM_CPP_API_LOG_LEVEL` is the odd one out in this package's environment family. Everything else a user sets is `IQM_BASE_URL`, `IQM_QC_ID`, `IQM_QC_ALIAS`, `IQM_TOKEN`, `IQM_TOKENS_FILE`, `IQM_JOBS_DIR` — and then the log level carries a `CPP_API` infix that describes an internal layer rather than anything the user picked. It is also the one variable the SPANK plugin does not inject, so a user on a Slurm cluster has no documented way to raise the log level of a job they just submitted. Those two facts compound: the variable is both hard to guess and unreachable where you most want it.

`IQM_LOG_LEVEL` becomes the primary name. `IQM_CPP_API_LOG_LEVEL` still works and is read only when the new name is unset or empty, so nothing that sets it today changes behavior. Empty counts as unset on purpose: schedulers and container runtimes export empty variables routinely, and since a name the parser does not recognize selects `NONE`, an exported-but-empty `IQM_LOG_LEVEL` would otherwise disable logging outright *and* hide a correctly set `IQM_CPP_API_LOG_LEVEL` — precisely the migration case the alias exists to cover. Reading the deprecated name now also logs a notice, so the alias does not quietly outlive its deprecation.

## The enabling change

Level resolution moved out of the `Logger` constructor into `Logger::level_from_environment()`. The singleton reads its level once per process, so the old code could not be exercised from a single test binary — the first test to touch the logger fixed the level for every test after it. Pulling the resolution into a static function makes the precedence rule directly testable, which is what the new tests in `test/unit/test_logging.cpp` do.

The deprecation notice went the same way, for the same reason: it started out inside the constructor, where a coverage run that sets the current variable never reaches it. `Logger::deprecation_notice()` decides and returns the text, the constructor only emits it, and the decision is covered directly.

## The SPANK side

The plugin now maps `iqm_log_level` in `plugstack.conf` and `--iqm-log-level` on the command line to `IQM_LOG_LEVEL`, through the same generic option table the other variables already use. An administrator can set a site default and a user can override it per job.

The value is checked against the levels the library accepts rather than passed straight through. Every name the library does not recognize means "no logging at all", so `--iqm-log-level=Debug` would otherwise silence the compute node for a user who asked for the opposite — and there is no way to warn them afterwards, because warning requires logging. The option table gained an optional per-entry validator for this, so `--iqm-log-level` fails at submission with the accepted names, and an invalid `plugstack.conf` value warns and is ignored instead of taking effect.

**Not covered by an automated test.** The SPANK plugin is compile-checked here against the local Slurm headers; the smoke and Docker suites were not run.

## Breadth, and why

Renaming a variable that CI passes through means touching every place that passes it: seven workflow files, `pyproject.toml`'s `environment-pass` for the wheel builds, `.env.example`, `docs/usage.md`, and `docs/spank_plugin.md`. That is mechanical, but it is most of the diff, and it is why this is worth reviewing on its own rather than alongside four unrelated fixes.

`docs/usage.md` also gains a warning that `DEBUG` logs raw request and response bodies. That was already true and undocumented; making the level easier to reach makes it worth saying out loud.

## Testing

Full unit suite green at 140 tests with `main` merged in. `uvx nox -s lint` clean. `clang-tidy` clean on every changed C++ file including `spank/iqm_spank_plugin.cpp`. Live integration tests were not run.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -48,6 +48,16 @@
 #IQM_TOKENS_FILE=
 
 # -------------------------------------------
+# Logging
+# -------------------------------------------
+
+# Verbosity of the QDMI device log, written to stderr
+# One of NONE, ERROR, INFO, DEBUG; any other value disables logging
+# DEBUG logs raw request and response bodies, so keep it off shared logs
+# Default: ERROR
+#IQM_LOG_LEVEL=
+
+# -------------------------------------------
 # Offloader Configuration
 # -------------------------------------------
 
--- a/.github/workflows/cpp-coverage.yml
+++ b/.github/workflows/cpp-coverage.yml
@@ -75,7 +75,7 @@ jobs:
           IQM_BASE_URL: "https://resonance.iqm.tech"
           IQM_QC_ALIAS: ${{ inputs.iqm-qc-alias }}
           IQM_TOKEN: ${{ secrets.IQM_TOKEN }}
-          IQM_CPP_API_LOG_LEVEL: "DEBUG"
+          IQM_LOG_LEVEL: "DEBUG"
         run: ctest -C Debug --output-on-failure --test-dir build --timeout 600 --repeat until-pass:3
 
       - name: Upload coverage to Codecov
--- a/.github/workflows/cpp-install-tests.yml
+++ b/.github/workflows/cpp-install-tests.yml
@@ -92,5 +92,5 @@ jobs:
         env:
           IQM_QC_ALIAS: ${{ inputs.iqm-qc-alias }}
           IQM_TOKEN: ${{ secrets.IQM_TOKEN }}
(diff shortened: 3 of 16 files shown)


burgholzer (approved):
This also feels fairly clean to me and can happily go in once GitHub is happy. 😃

PR #213, opened 2026-08-25
🩹 Fail a job whose submission response carries no job ID

marcelwa:
🤖 *AI text below* 🤖

`Handle_response` reports success only for HTTP 2xx. So by the time a submit path parses the submission response, the IQM Server has accepted the request and a job is queued remotely. If that response then turns out to be unreadable — not JSON, or JSON without a string `id` — this client has lost the job ID, and with it every way to poll, cancel, or collect that job.

What happened next is the part worth looking at. `job->status_` was left at `QDMI_JOB_STATUS_CREATED`, which is exactly the state `IQM_QDMI_device_job_submit` requires, so the same handle accepted a second submission. A caller doing the natural thing on `QDMI_ERROR_FATAL` — retry — queues a duplicate job onto the hardware while the first one still runs untracked. On a shared, rate-limited quantum computer where submission is the most expensive request there is, that is the wrong direction to fail in.

This is a deliberate reversal, not an oversight being corrected. `JobSubmissionRejectsResponsesWithoutJobId` pinned the old behaviour explicitly:

```cpp
// An unreadable submission response says nothing about the remote job, so
// the handle stays usable and can be submitted again.
```

That premise does not hold at that point in the code. A 2xx *is* the server saying it took the job; the ambiguity the comment describes belongs to the branch above it, where `Handle_response` failed and the request may never have arrived. The two cases were treated as one. This PR splits them: a transport or HTTP failure keeps failing the job as before, and an accepted-but-unreadable submission now fails it too, for the opposite reason — not because the job probably does not exist, but because it probably does and cannot be reached.

If you would rather keep the handle retryable, the alternative is to say so in the docs and accept the duplicate-submission risk; I did not take that route because the caller has no way to detect the orphan and nothing to cancel it with. Happy to flip it if you read the trade-off the other way.

Also here: the job ID is now read through a `find` and an `is_string()` check rather than `at("id")`, so a wrongly typed ID is handled where it happens instead of unwinding into the entry point's catch-all.

### Tests

- `JobSubmissionRejectsResponsesWithoutJobId` drives a fresh job per malformed response and asserts each reports `QDMI_JOB_STATUS_FAILED` and refuses a retry with `QDMI_ERROR_BADSTATE`.
- `CalibrationSubmissionWithoutAJobIdFailsTheJob` covers the calibration submit path, which reads its response the same way.
- `CalibrationResultsRejectMalformedStatusResponses` opened by submitting an unreadable response and then resubmitting the same handle. That is incidental to the malformed *status* responses it exists to cover, and it relied on the behaviour being changed here, so it is gone.

141 unit tests pass locally.

### Note for #200

`IQM_QDMI_device_job_submit_run` on the pulse-level branch copies this same shape, so #200 will want the same treatment once whichever of the two lands second catches up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,7 @@ releases may include breaking changes.
 ### Fixed
 
 - 🩹 Fail a job whose submission the IQM Server accepted but whose response
-  carried no readable job ID, instead of leaving the handle submittable. A retry
-  would have queued a second job while the first ran untracked ([#213])
+  carried no readable job ID, instead of leaving the handle submittable ([#213])
   ([**@marcelwa**])
 
 ## [1.4.0] - 2026-08-25
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ releases may include breaking changes.
 
 ## [Unreleased]
 
+### Fixed
+
+- 🩹 Fail a job whose submission the IQM Server accepted but whose response
+  carried no readable job ID, instead of leaving the handle submittable. A retry
+  would have queued a second job while the first ran untracked ([#213])
+  ([**@marcelwa**])
+
 ## [1.4.0] - 2026-08-25
 
 ### Added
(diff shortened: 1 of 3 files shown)


burgholzer (changes requested):
I am honestly not sure about changes like these. This change touches roughly 100 lines for something that is hypothetical at best. If the Server API suddenly doesn't report back a Job ID anymore, something way more fundamental is broken in the integration and this also does not feel like something to recover from.
The description of this PR also triggers me quite a bit because it is so overly verbose. This could have been a two-liner description and it would have likely been equally expressive.
Sorry for being so skeptical about these changes. I am just not quite sure whether they really help. In my opinion the code has become harder to read because of these changes.
Open to discuss.

marcelwa:
I'm closing this after our discussion offline.

PR #214, opened 2026-08-25
✨ Slow down before the IQM rate limit blocks the device

### LINKED ISSUE #198: ✨ Slow down before the rate limit blocks us, using the quota headers the API already sends
🤖 *AI text below* 🤖

## Problem Statement

The device reacts to rate limiting only after it has already been blocked. `Perform_with_retries` in `src/internal/http_client.cpp` retries on HTTP 429 and obeys the `Retry-After` header, which is correct but purely reactive: by the time it runs, the account has been blocked for 30 seconds and the damage is done.

The IQM Server API publishes the information needed to avoid that entirely. Every successful response carries `RateLimit-Limit` and `RateLimit-Remaining`, and the API documentation is explicit about how it expects clients to use them:

> Clients should monitor `RateLimit-Remaining` and slow down before it reaches zero, rather than trying to derive an absolute request rate from these values.

This device never reads either header.

The quota is worth spelling out, because it makes clear how easy it is to exhaust:

- Quota: **2000 units**, over a **10-second** rolling window
- Block on exhaustion: **30 seconds**
- Job cancellation: **100** units
- Job submission: **100** units flat, plus 100 for every 10 MiB of payload
- Status queries, property queries, and most reads: **10** units

So twenty job submissions inside one window exhaust the quota, and each one is worth ten status polls. A 30-second block is long enough to change the answer to the question that triggered it: a mock job finishes in well under a second, so a blocked status query returns after the job has left the queue, and a blocked cancellation returns after the job can no longer be cancelled.

That is not hypothetical. It is the cause of two separate integration-test flakes — #196 for `QueuePropertiesOnGarnetMock`, #197 for `JobCancellation` — and both were fixed by making the tests tolerate the outcome, because neither could be fixed in the layer where the problem actually is. The consequences are worse outside the test suite: the SPANK plugin and the offloader submit and poll on behalf of real workloads, where a 30-second stall is a scheduling problem rather than a red tick.

## Proposed Solution

Track the remaining quota from the response headers and slow down before it reaches zero, rather than waiting to be blocked.

- Record `RateLimit-Limit` and `RateLimit-Remaining` from every successful response.
- Before issuing a request, if the last observed remaining quota is below a threshold, wait out the remainder of the window instead of spending the request. Even a short wait is far cheaper than a 30-second block.
- Keep the existing 429 and `Retry-After` handling exactly as it is. This is a complement, not a replacement — the headers can be missing, stale, or wrong, other clients share the same account quota, and the reactive path stays the safety net.
- Log at debug when throttling engages, so a slow run is diagnosable rather than mysterious.

Some design notes from reading the current code:

- `Perform_with_retries` is a single choke point that both `Get` and `Post` route through, so the state and the decision have one obvious home and no call site needs to change.
- `Get_hooks().sleep` is already injectable and already stubbed by the unit tests, so this is testable hermetically without any real waiting — the existing rate-limit tests show the pattern.
- The quota is per user account, not per session, so the state should be shared across sessions in the process and guarded for concurrent use. It cannot be complete: another process or another client using the same token spends from the same quota, which is another reason to keep the reactive path.
- The headers appear on successful responses, so the tracker learns nothing from a request that fails for other reasons. Treating an unknown quota as "proceed" keeps the behavior no worse than today.

Worth deciding explicitly: whether the threshold and the whole behavior should be configurable, and whether it should be on by default. Being on by default is what makes the flakes go away; a way to turn it off is useful for anyone who would rather take the 429.

marcelwa:
🤖 *AI text below* 🤖

Closes #198.

The device only reacted to rate limiting once it was already blocked for 30 seconds. The IQM Server API reports the remaining quota on every successful response and asks clients to slow down before it runs out, and this device read neither header.

A session now keeps what its own responses reported and, once the quota falls below ten percent of the limit, waits out the rest of the ten-second window before its next request. The wait comes out of that request's timeout, and a request with less time left goes out unthrottled. `IQM_RATE_LIMIT_THRESHOLD_PERCENT` moves the threshold or turns it off. `Retry-After` is untouched and stays the safety net, since other clients spend from the same account quota.

Rewritten since the review: the quota moved off a process-wide static behind a mutex onto the session, which took most of the machinery with it. Down from +840/-83 to +392/-36, and from 24 tests to 8.


marcelwa:
🤖 *AI text below* 🤖

Adversarial review found three defects, now fixed.

- `limit * threshold_percent` overflowed for a `RateLimit-Limit` above `INT64_MAX / 100` and wrapped negative, disabling the wait exactly when the quota read empty. Confirmed under UBSan; the reported limit is now bounded.
- `Rate_limit_wait` cleared the observation before the caller could decline it, so a poll on a short timeout swallowed the reading and the next submission on the default timeout went out blind. The same one-shot flag let one thread wait while the rest stampeded. It is now a pure query for how much of the window is left, which is simpler and fixes both.
- Resolving the threshold happened under the tracker mutex, serialising every request on an environment lookup. Hoisted out.

The tests were the weaker half. `HttpStub` stubbed the sleep but not the clock, so a wait computed from the elapsed window was indistinguishable from a hardcoded ten seconds and the assertions would have flaked on a loaded runner. A clock hook joins the seam. Mutation testing drove the new cases: six changes, including deleting the elapsed-time computation outright, previously left all 151 tests green. All are caught now, at 164.

Not acted on: rounding the wait up past a sub-ten-second budget is the budget check working, and the default timeout is an hour. That interaction is now spelled out on the `timeout` parameter.

src/internal/http_client.cpp
```diff
@@ -112,27 +115,192 @@ void Default_sleep(const int seconds) {
...
+  std::int64_t remaining = 0;
+  std::int64_t limit = 0;
+  std::chrono::steady_clock::time_point observed_at;
+  bool known = false;
+};
+
+Rate_limit_tracker &Get_rate_limit_tracker() {
+  static Rate_limit_tracker tracker;
+  return tracker;
+}
```
burgholzer: This is likely me being stupid but does it really make sense to have this globally static? This means the same rate tracker is used for two concurrent sessions using different credentials and devices. That feels wrong to me.
This should be scoped to the session in my opinion, even though it may not always be perfectly accurate for various reasons.
This feels like a large synchronisation bottleneck for comparatively little value.

burgholzer (changes requested):
I get the idea, but this PR feels like triple the size that it needs to be to achieve its goal. It is overly verbose and feels bloated in several aspects. Please try to trim this down.

src/internal/http_client.cpp
marcelwa: 🤖 *AI text below* 🤖

Agreed on both counts, and the global was doing worse than you describe — it also meant an unrelated session could make a request wait for a quota it never spent from. It is gone. The reading is now a plain `Rate_limit_budget` aggregate on `IQM_QDMI_Device_Session_impl_d`, passed to `Get`/`Post` as a pointer that a caller without a session leaves null, and it carries no lock at all: nothing else on the session synchronizes either, and a mutex member would have cost the `noexcept` move that `IQM_QDMI_device_session_init` static-asserts on.

You are right that this is not perfectly accurate — the quota is per account, so another process using the same token spends from the same pool. That is why the reactive `Retry-After` path stays untouched as the safety net.

Dropping the shared tracker took a lot of machinery with it, which is most of the trim you asked for: +431/-43 now, from +840/-83, and 8 tests instead of 24.

marcelwa:
I trimmed down the PR description and the code. I would be grateful for another round of feedback.

burgholzer (approved):
This is so much cleaner now! Great work! 🙌
I have no further complaints. Feel free to merge whenever you are happy as well!

PR #218, opened 2026-08-26
⬆️ Require compatible Qiskit releases

burgholzer:
🤖 *AI text below* 🤖

## Description

Raise the direct Qiskit floor to 2.0 on Python 3.10–3.13 and 2.1 on Python 3.14 and newer.

`qiskit-algorithms` 0.4 cannot import against the previous Qiskit 1.1 floor. Qiskit 1.3 fixes that import but still crashes while transpiling against MQT Core 3.9.1 global target operations. Qiskit 2.0 is the first tested release where the complete minimum suite passes. Python 3.14 needs Qiskit 2.1 because the older dependency set has no usable `symengine` wheel.

## Validation

- `uvx nox -s lint`
- `uvx nox -s minimums` (Python 3.10–3.14: 58 passed and 7 credential-gated skips per version)

AI assistance: Codex isolated both dependency failures, found the first passing floors, and validated the full minimum matrix under human direction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ releases may include breaking changes.
   `IQM_SLURM_PARTITION`, so a site whose quantum partition is not named
   `quantum` can use the offloader ([#202]) ([**@marcelwa**])
 
+### Fixed
+
+- ⬆️ Require Qiskit 2.0 on Python 3.10–3.13 and Qiskit 2.1 on Python 3.14 and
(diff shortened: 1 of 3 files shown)


marcelwa:
@burgholzer the new Qiskit floor has to be documented in `docs/dependencies.md`.

burgholzer:
> @burgholzer the new Qiskit floor has to be documented in `docs/dependencies.md`.

Good catch. I almost figured that this went in too smoothly.

PR #219, opened 2026-08-26
🩹 Handle QDMI jobs that omit the shot count

marcelwa:
🤖 *AI text below* 🤖

MQT Core 3.9.2 lets a client submit a QDMI job without ever setting `SHOTSNUM` (munich-quantum-toolkit/core#2258), which makes a state this device never had to handle reachable: it kept `size_t num_shots_ = 0` and wrote `json_program["shots"]` unconditionally, so an omitted count would have asked the IQM Server to run zero shots. The count is now a `std::optional`, a circuit job carrying none is rejected with `QDMI_ERROR_INVALIDARGUMENT` before any request goes out and stays submittable once the count arrives, and `QDMI_DEVICE_JOB_PROPERTY_SHOTSNUM` answers `QDMI_ERROR_NOTSUPPORTED` while it is unset. Calibration jobs already arrived without a shot count and are otherwise unaffected, though `Job::getNumShots` now throws for them instead of returning a fabricated zero — say so if you would rather keep the zero. Eleven existing unit tests submitted circuit jobs without a shot count and now set one; two new ones cover the rejection and the property query, and there is no Python test because every job-submitting path here needs a live Resonance session. Once #200 lands, pulse jobs in `QDMI_PROGRAM_FORMAT_CUSTOM1` want the opposite treatment — the `RunDefinition` payload owns its repetition count, so that format should reject `SHOTSNUM` outright — but that belongs in its own PR. Verified with a Release Ninja build: 145 unit tests pass, `clang-tidy` is clean on both changed C++ files, `uvx nox -s lint` is green, and `pytest test/python` is 59 passed / 7 skipped (all live-credential).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ releases may include breaking changes.
 
 ### Fixed
 
+- 🩹 Reject a circuit job that carries no shot count instead of asking the IQM
+  Server to run zero shots, and report an unset shot count as absent rather than
+  as zero, now that MQT Core 3.9.2 lets a client omit it ([#219])
+  ([**@marcelwa**])
 - ⬆️ Require Qiskit 2.0 on Python 3.10–3.13 and Qiskit 2.1 on Python 3.14 and
   newer so the supported minimum environments install and run ([#218])
   ([**@burgholzer**])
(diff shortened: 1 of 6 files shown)


burgholzer (approved):
There's a bit of an overlap between this PR and the work that just landed as part of #200. I'll let you be the judge whether you'd still like to merge this (the changes here look fine to me 👍🏼) or whether #200 already covers most of the functionality added here.
I am pretty much done for the day when it comes to the other PR.

marcelwa:
> There's a bit of an overlap between this PR and the work that just landed as part of #200. I'll let you be the judge whether you'd still like to merge this (the changes here look fine to me 👍🏼) or whether #200 already covers most of the functionality added here.

Ah yes, there is quite some overlap. No worries though. Lets get #200 in first and then see what survives here.

> I am pretty much done for the day when it comes to the other PR.

Thanks for your work!

PR #220, opened 2026-08-28
💥 Drop support for Python 3.10 and x86 macOS

denialhaag:
🤖 *AI text below* 🤖

Drop support for Python 3.10 and x86 macOS in line with MQT Core. Require Python 3.11, use Python 3.11 as the stable ABI baseline, raise the macOS deployment target to 13.3, stop testing on `macos-26-intel`, stop publishing x86 macOS wheels, and align the inline-script and contributor metadata.

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -25,7 +25,6 @@ jobs:
           [
             ubuntu-24.04,
             ubuntu-24.04-arm,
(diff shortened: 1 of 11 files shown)


burgholzer (approved):
Thanks @denialhaag 🙏🏼 This will fit nicely in the next minor release!

PR #227, opened 2026-09-01
🔒️ Stop the SPANK plugin from exporting its bundled dependencies into slurmstepd

marcelwa:
🤖 *AI text below* 🤖

The SPANK plugin links `iqm::qdmi_device_internals`, which attaches cpr, libcurl, zlib, and OpenSSL as PUBLIC on an OBJECT library, so those symbols end up in the dlopened module with default visibility and are candidates for interposition inside `slurmstepd` alongside other site plugins — 576 exported symbols where the SPANK ABI needs 10. This adds a linker version script that exports exactly the four metadata symbols and six hooks the plugin defines and hides the rest, plus an assertion in the Docker test run so a dropped version script fails instead of silently widening the surface again. Nothing observable changes for users; `--version-script` is GNU ld/lld only, which is fine because the plugin requires `slurm/spank.h` and is built solely on Linux in CI.

### DIFF
--- a/spank/CMakeLists.txt
+++ b/spank/CMakeLists.txt
@@ -26,6 +26,18 @@ target_link_libraries(
                                     qdmi::qdmi_project_warnings)
 set_target_properties(${BUILD_IQM_SPANK_TARGET} PROPERTIES PREFIX "")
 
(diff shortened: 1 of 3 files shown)


burgholzer (approved):
This makes a lot of sense and feels like a clean solution 👍🏼

PR #228, opened 2026-09-01
✅ Let CTest select the offline test suite

marcelwa:
🤖 *AI text below* 🤖

The integration tests reach a live IQM backend, but they registered with CTest without a label, so selecting the offline suite meant knowing the build layout and pointing ctest at `build/test/unit`. They now carry an `integration` label, which makes `ctest -LE integration` run everything that works without credentials and `ctest -L integration` run the live suite, from the top-level build directory. Default behaviour is unchanged and every CI workflow keeps running the whole suite; note that the label has to precede `DISCOVERY_TIMEOUT` in the argument list, since that keyword ends the `PROPERTIES` list and anything after it is dropped silently.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -259,6 +259,10 @@ ctest -C Release --test-dir build --output-on-failure
 ctest -C Release --test-dir build/test/unit --output-on-failure
 ```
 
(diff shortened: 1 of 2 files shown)


burgholzer (approved):
LGTM 👍🏼

PR #229, opened 2026-09-01
✨ Answer the standard needs-calibration device query

marcelwa:
🤖 *AI text below* 🤖

`QDMI_DEVICE_PROPERTY_NEEDSCALIBRATION` was unhandled, so a client asking whether the quantum computer needs calibration got `QDMI_ERROR_NOTSUPPORTED` instead of an answer. The QDMI contract makes it a `size_t` where zero means no calibration is needed, and the IQM Server publishes no field carrying that meaning — health reports `{healthy, updated_at}`, the queue endpoint reports length and availability, and the calibration-set endpoints only describe the current default set — so with recalibration scheduled by IQM and an unfit system already surfacing as `QDMI_DEVICE_STATUS_MAINTENANCE`, zero is the honest answer, matching QDMI's own reference device and its conformance test. Happy to derive it from a server signal instead if one exists that I could not find in the API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ releases may include breaking changes.
 
 ### Added
 
+- ✨ Answer `QDMI_DEVICE_PROPERTY_NEEDSCALIBRATION` with zero instead of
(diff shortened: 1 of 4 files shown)


burgholzer (approved):
LGTM 👍🏼

PR #230, opened 2026-09-01
🔧 Round out the README badges and the packaging metadata

marcelwa:
🤖 *AI text below* 🤖

The README advertised only static facts, so nothing there told a reader whether the published package was current, whether main was building, or what the coverage gate said, and the package metadata pointed everywhere except at the documentation site. The sdist also shipped the C++ and Python test trees and the SPANK plugin — 40 of its 79 entries — even though a source install configures with `BUILD_IQM_QDMI_TESTS=OFF` and `BUILD_IQM_SPANK` off, and `ty` exited zero on warning-level diagnostics, which `[tool.ty.terminal]` now turns into a failure. A wheel still builds from the trimmed sdist, `check-sdist` reads the exclude list and stays quiet, and the shields URL spells `CI.yml` exactly as the workflow file does, since shields matches it case-sensitively.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,10 +9,13 @@
 
 # QDMI on IQM
 
+[![PyPI](https://img.shields.io/pypi/v/iqm-qdmi?logo=pypi&style=flat-square)](https://pypi.org/project/iqm-qdmi/)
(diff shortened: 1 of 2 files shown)


burgholzer (approved):
LGTM :shipit:

PR #231, opened 2026-09-01
🩹 Fix IQM JSON payloads and harden offloaded results

marcelwa:
🤖 *AI text below* 🤖

IQM submission still used legacy PRX arguments in turns, while the current protocol requires radians. The Slurm offloader also loaded worker-controlled pickle results. This change corrects the input payload, uses validated JSON results, and isolates live-backend CI failures.

- Emit PRX `angle` and `phase` in radians and update Python/C++ fixtures without changing their rotations. Log server job failure messages at ERROR level.
- Preserve joint sampling counts across registers; validate binary keys and integer counts. Validate finite VQE values, dimensions, and optimizer counters before reconstructing `VQEResult` with the original ansatz. Remove obsolete pickle compatibility code.
- Keep wheel tests and C++ platform, installation, coverage, and sanitizer checks offline. Retain a separate required C++ integration matrix for Emerald, Garnet, and Sirius, without retrying outages.
- Use Garnet for the seven generic live Python checks and executed Qiskit documentation. Bound Python job waits to 120 seconds and register cancellation before waiting, including primitive submissions.
- Cancel unfinished C++ integration jobs and free their handles during fixture teardown, including fatal assertions and skips. Update documentation and changelog.

No `iqm-client` dependency is added to the package, tests, lockfile, or CI.

Validation at `c3d742c`:

- 97 offline Python tests and all 7 live Python tests on Garnet passed.
- All 162 offline C++ tests passed, including a full ASan/UBSan run. Garnet integration passed (9 tests passed, optional calibration skipped, QIR disabled as before).
- Three hermetic LeakSanitizer regression checks cover timeout/assertion cleanup, cancellation/status failures, and created/terminal/null handles.
- Full lint and whole-file clang-tidy 22 passed.
- Executed documentation build passed, including the Qiskit notebook on Garnet; 14 reference-resolution warnings remain.
- Isolated stock `iqm-client[pin-iqm]==35.0.3` validates the updated payload and independently reproduces the Resonance problem without QDMI.

Remaining external failure: Resonance reports Emerald and Sirius `healthy=false` despite `operational_status=online`; Garnet is healthy and completes. A stock SDK Sirius job (`01a0bb71-2f56-75e2-8ea3-df04484f077b`) failed after about 268 seconds with `Could not send job to the QC`, source `iqm-station-control`, error code `unknown`. An equivalent Garnet job completed. Emerald remained queued for 900 seconds and was then cancelled; cancellation was confirmed. The final 21:07 UTC health snapshot was unchanged. This leaves the Emerald/Sirius backend service as a remaining blocker; their integration checks remain required. The hosted rerun at `c3d742c` is complete: 34 checks passed, including documentation, sanitizers, coverage, all platform/installation checks, and every wheel build. Only the required Emerald/Sirius integration checks and their aggregate gate failed. Emerald's logs confirm both timed-out jobs were cancelled during teardown; Sirius reports the same `Could not send job to the QC` error at ERROR level.

[CI run](https://github.com/iqm-finland/QDMI-on-IQM/actions/runs/35469353565), [documentation run](https://github.com/iqm-finland/QDMI-on-IQM/actions/runs/35469353118).


marcelwa:
🤖 *AI text below* 🤖

Reviewed this adversarially and pushed the fixes onto the branch rather than leaving them as comments. Two holes in the guard: `pickle` resolves a *dotted* global name by walking attributes out of the named module, so an allowed module paired with a dotted path reached whatever that module had imported, and `getattr` was reaching any non-dunder attribute of an allowed class, including methods that write a file of the payload's choosing — `find_class` now accepts only a plain module-level global that is also a class, `getattr` only the three reconstruction hooks the bit and parameter types actually reduce through, and `qiskit.capi` is excluded outright since it wraps the native library rather than any result type. Separately, the allowlist was too tight to load a real result: walking the opcodes of genuine `sample`/`estimate` payloads on Qiskit 2.0, 2.1 and 2.5 shows they also need `functools.partial`, `uuid.UUID` and (on 2.0) symengine's `load_basic`, so `estimate` refused every ansatz beyond a single parameterized gate — the round-trip test's ansatz happened to be the one shape that avoided all three, and now uses a parameter vector and a two-qubit gate. Each restriction has exactly one test case that goes red when that restriction alone is removed, and restoring `pickle.loads` still turns all five refusal cases red.

burgholzer (approved):
@marcelwa Thanks for kickstarting this.
I played around with the PR quite a bit; especially after the IQM mocks on Resonance were failing until very recently due to some unrelated errors.

I had the impression that the JSON logic for result handling was very brittle and basically replicated Qiskit internals, which are subject to change. As such, I reverted back to the pickle implementation for now. This is executed in a controlled environment where I am not concerned with the typical unsafety of pickle.

I believe the remaining fixes here are pretty reasonable and we can happily get them in. I'll leave it to you whether to simply merge this or whether you want to add some further changes.

marcelwa:
> @marcelwa Thanks for kickstarting this. I played around with the PR quite a bit; especially after the IQM mocks on Resonance were failing until very recently due to some unrelated errors.
> 
> I had the impression that the JSON logic for result handling was very brittle and basically replicated Qiskit internals, which are subject to change. As such, I reverted back to the pickle implementation for now. This is executed in a controlled environment where I am not concerned with the typical unsafety of pickle.
> 
> I believe the remaining fixes here are pretty reasonable and we can happily get them in. I'll leave it to you whether to simply merge this or whether you want to add some further changes.

Thanks @burgholzer! I can live with the current state. Still a bit worried about `pickle`, but I guess you're right: it is a controlled environment after all. I just fixed the CHANGELOG conflict and will merge once the CI is green.

PR #232, opened 2026-09-01
🐛 Decode base64url JWT payloads

marcelwa:
🤖 *AI text below* 🤖

JWT payloads are base64url-encoded per RFC 7515, but the decode table in `src/internal/iqm_auth.cpp` only mapped the standard `+` and `/`, leaving `-` and `_` unmapped and silently decoded as garbage. A payload byte equal to `>`, `?`, `~`, `DEL`, or non-ASCII landing at an offset congruent to 2 (mod 3) is enough to trigger it, and the usual outcome is that `time_left_seconds` cannot parse the payload and reports a perfectly valid token as expired. The table now maps both alphabets, and four unit tests cover the two base64url characters, the unpadded body lengths RFC 7515 produces, and a body length that cannot be base64 at all.


burgholzer (approved):
LGTM 👍🏼 Nice catch

PR #242, opened 2026-09-05
🐛 Link MinGW unit tests

### LINKED ISSUE #241: 🐛 Link unit tests correctly with MinGW
🤖 *AI text below* 🤖

## System and Environment Information

- Operating system: Windows 10.0.19045.7663
- IQM QDMI Device: main at 5dfab69
- Compiler: MinGW-w64 GCC 16.2.0
- CMake: 4.4.2
- Generator: Ninja 1.13.2

## Bug Description

The documented CMake build compiles the IQM QDMI implementation through the iqm_qdmi_device_internals object library and includes those objects directly in the unit-test executable. On Windows, however, the generated QDMI declarations seen by the unit-test sources use dllimport. MinGW therefore emits references such as __imp_IQM_QDMI_device_session_init, even though the implementation is linked directly into the same executable rather than imported from a DLL.

The unit_tests link fails with undefined references to __imp_IQM_QDMI_device_session_init and the other IQM_QDMI device entry points.

The object-library implementation itself is correctly compiled with IQM_QDMI_device_EXPORTS. Its Windows consumers also need IQM_QDMI_STATIC_DEFINE because they consume the object files directly.

## Steps to Reproduce

1. On Windows with MinGW-w64 GCC and Ninja, check out main at 5dfab69.
2. Configure with: cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DBUILD_IQM_QDMI_TESTS=ON
3. Build with: cmake --build build --config Release --target unit_tests
4. Observe undefined references to __imp_IQM_QDMI_device_* while linking test/unit/unit_tests.exe.

Expected: the documented CMake build links the unit-test executable successfully with MinGW.

AI assistance: Codex reproduced the linker failure, traced the import decoration through the generated header and CMake usage requirements, and validated a local fix. A human maintainer remains responsible for accepting the report and fix.

marcelwa:
🤖 *AI text below* 🤖

## Description

Mark Windows consumers of the internal QDMI object library as static-definition users. The unit tests consume the implementation objects directly, so their generated QDMI declarations must not import entry points from a separate DLL.

Add a Windows compile-time guard to keep that CMake usage requirement covered.

Fixes #241.

## Dependency

This branch includes the token-fixture fix from #240 while that PR is pending, allowing the full CI matrix to validate this fix against a green baseline. Once #240 lands on main, its already-merged commit drops out of this PR diff.

## Validation

- Configured a Release MinGW-w64 GCC 16.2.0 build with unit tests enabled
- Built the complete unit_tests executable successfully; before this change it failed with undefined __imp_IQM_QDMI_device_* references
- All 157 offline unit tests passed
- uvx nox -s lint passed

AI assistance: Codex reproduced and diagnosed the import-decoration mismatch, prepared the CMake fix and compile-time guard, incorporated the independently reviewed #240 prerequisite, and ran the checks above. A human reviewer remains responsible for accepting the change.


burgholzer (approved):
Looks fine to me. Feel free to merge after resolving conflicts.

PR #246, opened 2026-09-05
:arrow_up: Adopt MQT Core version 3.10

marcelwa:
🤖 *AI text below* 🤖

Supersedes #245, which bumps MQT Core but leaves imports and estimator options that 3.10 removed. This migrates the examples and shared factories to native Qiskit primitives, with offline tests for ordered shots, counts-only estimation, and grouped observables. The default estimator now uses precision 1/64 (4,096 shots per measurement circuit); the QSCI example retains its explicit shot option.

### DIFF
--- a/test/python/test_qiskit_backend.py
+++ b/test/python/test_qiskit_backend.py
@@ -19,16 +19,23 @@
 
 from __future__ import annotations
 
+import json
 import os
 from pathlib import Path
 from typing import Any
+from unittest.mock import Mock
 
+import numpy as np
 import pytest
-from qiskit.circuit import QuantumCircuit
+from mqt.core.plugins.qiskit.backend import QDMIBackend
+from mqt.core.qdmi import Device, Job, ProgramFormat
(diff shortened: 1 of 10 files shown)


test/python/test_qiskit_backend.py
burgholzer, line 1: The tests here feel pretty overkill. They test public behavior of the Qiskit estimators as well as MQT Core, which each individually already test that functionality.
I would personally drop these.

burgholzer (approved):
Thanks @marcelwa 🙏 this looks largely good! Just one observation on some of the test additions. Feel free to adjust or ignore, then merge 😌

burgholzer (approved):
Edit: approved too soon. The above comment stil stands 🙃

test/python/test_qiskit_backend.py
marcelwa: Resolved now.

burgholzer (approved):
Now it really looks good to me. Let's get this in!

